PR #3, opened 2020-04-30
Adding duplicate command

JacobCallahan:
This allows us to duplicate the arguments used to checkout a
system in our current inventory.
Also fixed some typos.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Broker is a tool designed to provide a common interface between one or many serv
 ```
 cd <broker root directory>
 pip install .
-mv setting.yaml.example settings.yaml
+cp setting.yaml.example settings.yaml
 ```
 Then edit the settings.yaml file
 
@@ -20,11 +20,23 @@ broker checkout --workflow "workflow-name" --workflow-arg1 something --workflow-
 You can pass in any arbitrary arguments you want
 
 **Nicks**
+
 Broker allows you to define configurable nicknames for checking out vms. Just add yours to setting.yaml and call with the ```--nick``` option
 ```
 broker checkout --nick rhel7
 ```
+
+**Duplicating a VM**
+
+Broker offers another shortcut for checking out a VM with the same recipe as one already checked out by Broker. This is via the ```duplicate``` command.
+```
+broker duplicate my.awesome.vm.com
+broker duplicate 0
+broker duplicate 1 3
+```
+
 **Listing your VMs**
+
 Broker maintains a local inventory of the VMs you've checked out. You can see these with the ```inventory``` command.
 ```
 broker inventory
--- a/broker/commands.py
+++ b/broker/commands.py
@@ -7,6 +7,19 @@ from broker import logger as b_log
 from broker import helpers
 
 
+def _checkout(**broker_args):
+    """Internal function to provide a common checkout interface
+    This allows us to duplicate currently checked out systems
+    """
+    broker_inst = VMBroker(**broker_args)
+    broker_inst.checkout()
+    new_hosts = []
+    for host in broker_inst._hosts:
+        logger.info(f"{host.__class__.__name__}: {host.hostname}")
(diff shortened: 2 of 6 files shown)


JacobCallahan:
Wait for me to update the readme as well.

JacobCallahan:
@mshriver this should be good to go

README.md
```diff
@@ -20,11 +20,23 @@ broker checkout --workflow "workflow-name" --workflow-arg1 something --workflow-
 You can pass in any arbitrary arguments you want
 
 **Nicks**
+
 Broker allows you to define configurable nicknames for checking out vms. Just add yours to setting.yaml and call with the ```--nick``` option
 ```
 broker checkout --nick rhel7
 ```
+
+**Reproducing a VM**
```
kedark3, line 29: what are your thoughts around using `recreate` or `clone` ? `Reproduce` can be confusing to non-native English speakers IMHO.

JacobCallahan: recreate is a good alternative. I'm certainly not sold on the name. I'd like to avoid clone as this has a set meaning in a VM context.

broker/commands.py
```diff
@@ -64,10 +71,30 @@ def checkin(vm):
...
+@cli.command()
+@click.argument("vm", type=str, nargs=-1)
+def reproduce(vm):
+    """Reproduce a broker-procured vm
+
+    COMMAND: broker reproduce <vm hostname>|<local id>|all
+    """
+    inventory = helpers.load_inventory()
+    for num, host in enumerate(inventory):
+        if str(num) in vm or host["hostname"] in vm or "all" in vm:
```
omkarkhatavkar, line 92: `str(num) in vm ` consider here the num is entered is 1, we have more than 10 vm's e.g. 10 11 12 13 in that it would match any id containing that reproduce the instance. We should explicit ==?  
I'm assuming the VM is string type and not sorted list.

JacobCallahan: great catch!! you're absolutely right. i'll adjust, thanks!

JacobCallahan: ```vm```, in these cases, should always be a list (or tuple) can't remember which at this time. so the ```in``` membership check **shouldn't** have this issue. That said, I haven't tested that many in the inventory, so I can't say 100%.

omkarkhatavkar: Okay As I said earlier if the vm is collection object list tuple won't be an issue, I saw the `type=str` in parenthesis. So-called it.

README.md
mshriver: Yeah I'm not necessarily a fan of `reproduce` for this, but can't think of anything more appropriate that isn't `clone`.

Maybe `duplicate`?

broker/commands.py
mshriver: https://click.palletsprojects.com/en/7.x/arguments/#variadic-arguments

Always a tuple - doing some local testing, because I have some concerns about this statement as well.

mshriver: Logic is fine, but I'm thinking that `--all` should be a boolean arg here, not a special string for the vm arg.

README.md
kedark3: `duplicate` is better.

JacobCallahan: duplicate it is!

JacobCallahan:
@mshriver @kedark3 all updated

README.md
```diff
@@ -20,11 +20,23 @@ broker checkout --workflow "workflow-name" --workflow-arg1 something --workflow-
...
+
+Broker offers another shortcut for checking out a VM with the same recipe as one already checked out by Broker. This is via the ```reproduce``` command.
+```
+broker reproduce my.awesome.vm.com
```
kedark3, line 33: I think ReadMe needs to read `Duplicate` here.

JacobCallahan: yep! missed one. nice catch

PR #4, opened 2020-05-08
Documentation for helper.py

ColeHiggins2:
Documentation for helper functions in broker

### DIFF
--- a/broker/helpers.py
+++ b/broker/helpers.py
@@ -68,9 +68,9 @@ def flatten_dict(nested_dict, parent_key=""):
 def resolve_nick(nick):
     """Checks if the nickname exists. Used to define broker arguments
 
-    :param nick: pulled from dictionary by keyword
+    :param nick: String representing the name of a nick
 
-    :return: the value of NICKS
+    :return: a dictionary mapping argument names and values
     """
     nick_names = settings.get("NICKS", {})
     if nick in nick_names:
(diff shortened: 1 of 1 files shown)


broker/helpers.py
```diff
@@ -61,13 +66,22 @@ def flatten_dict(nested_dict, parent_key=""):
 
 
 def resolve_nick(nick):
-    """check if the nick exists and if so, return the arguments"""
+    """Checks if the nickname exists. Used to define broker arguments
+
+    :param nick: pulled from dictionary by keyword
```
JacobCallahan, line 71: this is specifically a string representing the name of a nick

broker/helpers.py
```diff
@@ -61,13 +66,22 @@ def flatten_dict(nested_dict, parent_key=""):
...
+
+    :return: the value of NICKS
```
JacobCallahan, line 73: how about instead of "the value of NICKS", "A dictionary mapping argument names and values"

broker/helpers.py
```diff
@@ -78,6 +92,14 @@ def load_inventory():
...
 
 def update_inventory(new_hosts, replace_all=False):
+    """Updates list of local hosts in the checkout interface
+
+    :param new_hosts: list of dictionaries
+
+    :param replace_all: True or False(Default False)
+
+    :return: no return value
+    """
```
JacobCallahan, line 102: this entire function is changing in #5, so we'll have to adjust after.

PR #5, opened 2020-05-08
Minor project refactor

JacobCallahan:
Changed the location of the logic for checkin and checkout
Consolidated most of the logic under the VMBroker class to allow
it to be more useful for non-cli users.
Added a TestProvider class in preparation for upcoming tests.

### DIFF
--- a/broker/broker.py
+++ b/broker/broker.py
@@ -1,10 +1,16 @@
 from logzero import logger
 from broker.providers.ansible_tower import AnsibleTower
+from broker.providers.test_provider import TestProvider
 from broker.hosts import Host
+from broker import helpers
 
-PROVIDERS = {
+
+PROVIDERS = {"AnsibleTower": AnsibleTower, "TestProvider": TestProvider}
+
+PROVIDER_ACTIONS = {
     # action: (InterfaceClass, "method_name")
-    "workflow": (AnsibleTower, "exec_workflow")
+    "workflow": (AnsibleTower, "exec_workflow"),
+    "test_action": (TestProvider, "test_action"),
 }
 
 HOST_CLASSES = {"host": Host}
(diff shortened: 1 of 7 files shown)


JacobCallahan:
I ran through broker's CLI functionality and everything I tried is working fine. Tests will be in progress after this refactor.

broker/broker.py
```diff
@@ -34,31 +45,51 @@ def _act(self, provider, method, checkout=False):
...
+            logger.info(f"Checking in {host.hostname}")
             host.release()
+            helpers.update_inventory(remove=host.hostname)
+
+    @staticmethod
+    def reconstruct_host(host_export_data):
+        """reconstruct a host from export data"""
+        logger.debug(f"reconstructing host with export: {host_export_data}")
+        provider = PROVIDERS[host_export_data.get("_broker_provider")]
+        if not provider:
```
omkarkhatavkar, line 76: `provider = PROVIDERS[host_export_data.get("_broker_provider")]` failing with keyerror incase on invalid provider
#### Result:
`KeyError: 'AnsibleTowerA' `
 
Below code see if it make sense?  
```
        if not host_export_data.get("_broker_provider") in PROVIDERS:
            logger.warning(
                f"No provider found with name {host_export_data.get('_broker_provider')}"
            )

            return
        provider = PROVIDERS[host_export_data.get("_broker_provider")]
```
#### Result: 
```
(env) [root@okhatavk broker]# broker checkin --all
[WARNING 200511 19:11:51] No provider found with name AnsibleTowerA
[WARNING 200511 19:11:51] No provider found with name AnsibleTowerA
[WARNING 200511 19:11:51] No provider found with name AnsibleTowerA

```

JacobCallahan: how did you get a provider named "AnsibleTowerA"?

omkarkhatavkar: I meant if someone gives the wrong the provider in the file, I meant that static code check, that if the condition will not be executed any time.

broker/commands.py
```diff
@@ -58,14 +47,11 @@ def checkin(vm, all_):
...
-        if str(num) in vm or host["hostname"] in vm or all_:
-            to_remove.append((num, host))
-    # reconstruct the hosts and call their release methhod
-    for num, host in to_remove[::-1]:
-        del inventory[num]
-        logger.info(f"Checking in {host['hostname']}")
-    helpers.update_inventory(inventory, replace_all=True)
+    for num, host_export in enumerate(inventory):
+        if str(num) in vm or host_export["hostname"] in vm or all_:
+            to_remove.append(VMBroker.reconstruct_host(host_export))
```
omkarkhatavkar, line 52: Do you think we need to raise here info message in case of invalid input VM's?

broker/broker.py
JacobCallahan: That is a file that should only be modified by broker, so we technically are good here. I will beef up line 75 though just in case. Thanks

broker/commands.py
JacobCallahan: The way this is written, you can supply any number of bad inputs and it simply won't act on them

PR #10, opened 2020-05-21
Update log level setting pattern

mshriver:
Small change to the log level setting pattern.

Needs testing, pushing to get early feedback on the approach.

### DIFF
--- a/broker/commands.py
+++ b/broker/commands.py
@@ -1,5 +1,6 @@
 from pathlib import Path
 import click
+import logging
 from logzero import logger
 from dynaconf import settings
 from broker.broker import VMBroker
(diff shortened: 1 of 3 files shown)


broker/commands.py
```diff
@@ -9,11 +9,11 @@
...
 @click.group()
-@click.option("--debug", is_flag=True)
-def cli(debug):
-    if debug or settings.DEBUG:
-        b_log.setup_logzero(level="debug")
-    logger.debug("I'm in debug mode")
+@click.option("--log-level",
+              type=click.Choice(['info', 'warning', 'error', 'critical', 'debug']))
+def cli(log_level):
+    b_log.update_logging_level(level=log_level)  # the method handles uppercase
```
JacobCallahan, line 15: should you instead just call setup_logzero as before?

broker/logger.py
```diff
@@ -6,27 +6,23 @@
...
+    formatter = logzero.LogFormatter(fmt=debug_fmt if log_level == logging.DEBUG else log_fmt)
     logzero.setup_default_logger(formatter=formatter)
     logzero.loglevel(level)
     logzero.logfile(
-        path, loglevel=level, maxBytes=1e9, backupCount=3, formatter=formatter
+        path, loglevel=log_level, maxBytes=1e9, backupCount=3, formatter=formatter
     )
 
 
+def update_logging_level(level="info"):
```
JacobCallahan, line 23: if you call setup_logzero, then this function wouldn't be needed

broker/commands.py
mshriver: Responding here as both comments are related.

Correct, but I considered this as a specific change to the log configuration that could be called separately.

In reviewing though I see that this won't end up updating the log formatting.

I'll switch up to re-call existing method, so the formatting gets updated

broker/commands.py
```diff
@@ -8,12 +9,23 @@
...
+
+
 @click.group()
-@click.option("--debug", is_flag=True)
-def cli(debug):
-    if debug or settings.DEBUG:
-        b_log.setup_logzero(level="debug")
-    logger.debug("I'm in debug mode")
+@click.option(
+    "--log",
```
JacobCallahan, line 20: a bit of a nitpick, but I think ```--log-level``` is a better option, as I would consider ```--log``` to potentially be a log path.

mshriver: You know I spent several minutes debating internally whether or not it was better shortened. :)

PR #11, opened 2020-05-21
Finish documentation

ColeHiggins2:
Documentation for commands.py and ansible_tower.py files.

### DIFF
--- a/broker/commands.py
+++ b/broker/commands.py
@@ -36,7 +36,15 @@ def cli():
 @click.pass_context
 def checkout(ctx, workflow, nick):
     """Checkout a Virtual Machine
-    COMMAND: broker checkout --<action> <argument>
+    COMMAND: broker checkout --workflow "workflow-name" --workflow-arg1 something
+    or
+    COMMAND: broker checkout --nick "nickname"
+
+    :param ctx: clicks content object
+
+    :param workflow: workflow template stored in AnsibleTower, passed in as a string
+
+    :param nick: shortcut for arguments saved in settings.yaml, passed in as a string
     """
     broker_args = {}
     if nick:
@@ -56,6 +64,10 @@ def checkin(vm, all_):
     """Checkin a VM or series of VMs
 
     COMMAND: broker checkin <vm hostname>|<local id>|all
+
+    :param vm:
+
+    :param all_:
     """
     inventory = helpers.load_inventory()
     to_remove = []
--- a/broker/commands.py
+++ b/broker/commands.py
@@ -35,14 +35,14 @@ def cli():
 @click.option("--nick", type=str, help="Use a nickname defined in your settings")
 @click.pass_context
 def checkout(ctx, workflow, nick):
-    """Checkout a Virtual Machine
+    """Checkout or "create" a Virtual Machine broker instance
     COMMAND: broker checkout --workflow "workflow-name" --workflow-arg1 something
     or
     COMMAND: broker checkout --nick "nickname"
 
     :param ctx: clicks content object
 
-    :param workflow: workflow template stored in AnsibleTower, passed in as a string
+    :param workflow: workflow template stored in Ansible Tower, passed in as a string
 
     :param nick: shortcut for arguments saved in settings.yaml, passed in as a string
     """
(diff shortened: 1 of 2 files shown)


broker/commands.py
```diff
@@ -23,8 +23,16 @@ def cli(debug):
...
 @click.pass_context
 def checkout(ctx, workflow, nick):
-    """Checkout a Virtual Machine
-    COMMAND: broker checkout --<action> <argument>
+    """Checkout or "create" a Virtual Machine broker instance
+    COMMAND: broker checkout --workflow "workflow-name" --workflow-arg1 something
+    or
+    COMMAND: broker checkout --nick "nickname"
+
+    :param ctx: clicks content object
```
mshriver, line 31: context*

broker/commands.py
```diff
@@ -57,7 +69,11 @@ def checkin(vm, all_):
 @cli.command()
 @click.option("--details", is_flag=True, help="Display all hist details")
 def inventory(details):
-    """Get a list of all VMs you've checked out"""
+    """Get a list of all VMs you've checked out showing hostname and local id
+        hostname pulled from list of dictionaries
+
+    :param details: click option to display all hist details
```
mshriver, line 75: host*

Its wrong in the click option help statement on L71 too.

broker/providers/ansible_tower.py
```diff
@@ -51,6 +51,14 @@ def _merge_artifacts(self, at_object, strategy="latest", artifacts=None):
            - latest: overwite existing values with newer values
            - branch: each branched child gets its own sub-dictionary (todo)
            - min-branch: only branch children if conflict is detected (todo)
+
+        :param at_object: object you want to merge
+
+        :param strategy: default to latest
```
mshriver, line 57: I think it would be good to actually move the text from L50-53 defining the strategies into this param block.  The default is defined by the function signature, its better to use these docblock statements to describe the parameter's purpose within the method.

broker/providers/ansible_tower.py
```diff
@@ -69,8 +77,20 @@ def _merge_artifacts(self, at_object, strategy="latest", artifacts=None):
...
+        :param provider_params: dictionary of what the provider returns when initially
+        creating the vm
+
+        :param host_classes: host object
+
+        :return: broker object of constructed host instance
+        """
         if provider_params:
             job = provider_params
+            print(job)
```
mshriver, line 91: Was this for debugging?  I don't think these provider methods should be printing, just logging.

broker/providers/ansible_tower.py
```diff
@@ -94,6 +114,12 @@ def construct_host(self, provider_params, host_classes, **kwargs):
         return host_inst
 
     def exec_workflow(self, **kwargs):
+        """Execute template job in Ansible Tower
+
+        :param kwargs: workflow template name passed in a string
+
+        :return: dictionary containing all information about executed workflow
```
mshriver, line 121: Is it really a dictionary?  We're calling `wait_until_completed()` on the returned name (`job`), and making attribute references.

broker/providers/ansible_tower.py
JacobCallahan: looks like one made it through! ha

broker/providers/ansible_tower.py
JacobCallahan: this is technically an object in awxkit, but the contents of the object are effectively a dictionary. 
an example of (a very reduced version of) the contents can be found here https://github.com/SatelliteQE/broker/blob/master/tests/data/ansible_tower/fake_jobs.json#L2

JacobCallahan:
@ColeHiggins2 any update on this?

broker/providers/ansible_tower.py
ColeHiggins2: yes, that was for debugging, must have missed it.

broker/providers/ansible_tower.py
ColeHiggins2: should I specify this in the documentation or is it fine as is?

JacobCallahan: it's fine enough.

PR #13, opened 2020-05-29
Add session functionality to Host objects

JacobCallahan:
Host objects can now establish an ssh connection to their corresponding host.
+ Added session class
+ Added session functionality to Host class
+ Updated VMBroker context manager
+ Added ssh2-python dependency (requires cmake)
+ New host settings added to settings.yaml.example

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,15 @@
 History
 =======
 
+0.0.7 (2020-05-29)
+==================
+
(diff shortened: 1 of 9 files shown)


JacobCallahan:
This change makes broker fully automation-ready!
Screenshot from 2020-05-29 16-30-02
Screenshot from 2020-05-29 16-34-00

PR #14, opened 2020-06-03
Add execute subcommand to broker

JacobCallahan:
+ Added execute functionality to broker
+ Added more functionality to VMBroker subclass to handle execute
+ Slightly changed AnsibleTower provider to allow for arbitrary workflow execution

### DIFF
--- a/broker/providers/ansible_tower.py
+++ b/broker/providers/ansible_tower.py
@@ -106,7 +106,7 @@ class AnsibleTower(Provider):
             logger.debug(f"hostname: {hostname}, name: {name}, host type: {host_type}")
             host_inst = host_classes[host_type](hostname=hostname, name=name, **kwargs)
         else:
-            host_inst = host_classes[kwargs.get('type')](**kwargs)
+            host_inst = host_classes[kwargs.get("type")](**kwargs)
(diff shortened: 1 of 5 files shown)


broker/providers/ansible_tower.py
```diff
@@ -118,15 +118,24 @@ def exec_workflow(self, **kwargs):
...
         job = wfjt.launch(payload={"extra_vars": str(kwargs).replace("--", "")})
         job.wait_until_completed(timeout=1800)
         if not job.status == "successful":
-            logger.error(f"Workflow Status: {job.status}\nExplanation: {job.job_explanation}")
+            logger.error(
+                f"Workflow Status: {job.status}\nExplanation: {job.job_explanation}"
+            )
             return
+        if artifacts:
+            return self._merge_artifacts(job)
```
kedark3, line 135: Are we merging artifacts from all nodes? or getting those from the last node in WF. I think the later is desired.

JacobCallahan: currently all. you will then be able to access the top level of what you're interested in with the jenkins export key.
If only the last node is desired, we can come up with a solution for that as well.

PR #15, opened 2020-06-09
Add Dockerfile

JacobCallahan:
Added ability to build broker in a container
You should mount the broker directory to /root/broker in the container

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM fedora
+MAINTAINER https://github.com/SatelliteQE
+
+RUN dnf -y install make cmake gcc-c++ zlib-devel \
+           openssl-devel git python3-pip python3-devel \
+           && dnf clean all
+WORKDIR /root/broker
+COPY . /root/broker/
+RUN pip install .
+RUN cp settings.yaml.example settings.yaml
+
+
+ENTRYPOINT ["broker"]
+CMD ["--help"]
--- a/broker/providers/ansible_tower.py
+++ b/broker/providers/ansible_tower.py
@@ -118,11 +118,10 @@ class AnsibleTower(Provider):
         :return: dictionary containing all information about executed workflow
         """
         workflow = kwargs.get("workflow")
+        artifacts = False
         if "artifacts" in kwargs:
             kwargs.pop("artifacts")
             artifacts = True
-        else:
-            artifacts = False
         wfjt = self.v2.workflow_job_templates.get(name=workflow).results.pop()
         job = wfjt.launch(payload={"extra_vars": str(kwargs).replace("--", "")})
         job.wait_until_completed(timeout=1800)


JacobCallahan:
Usage
```
podman run --privileged --rm -v $(pwd):/root/broker broker checkout --nick rhel7
podman run --privileged --rm -v $(pwd):/root/broker broker inventory
podman run --privileged --rm -v $(pwd):/root/broker broker checkin 1
```

Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM fedora
+MAINTAINER https://github.com/SatelliteQE
+
+RUN dnf -y install make cmake gcc-c++ zlib-devel \
```
kedark3, line 4: can we add a ```dnf clean all``` to clean out cache and stuff after install, helps keeping our image small.

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+           openssl-devel git python3-pip python3-devel
+RUN mkdir broker
```
kedark3, line 6: shall we use absolute path here? ```mkdir /root/broker```

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+COPY / /root/broker/
```
kedark3, line 7: why we copy `/` to broker?

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+WORKDIR /root/broker
+RUN cd /root/broker && pip install .
```
kedark3, line 9: I think `cd` here is redundant as we have `WORKDIR` already. Might wanna remove it and see what is the result.

Dockerfile
JacobCallahan: it's an option. if you don't mind, can you run a size comparison?

Dockerfile
JacobCallahan: we could, but shouldn't have to

Dockerfile
JacobCallahan: copying the contents of this repo to the directory, so it can be installed

Dockerfile
JacobCallahan: yeah this was originally above the workdir command. i can remove the cd

Dockerfile
kedark3: Sure I can help with that.

Dockerfile
kedark3: Please make it absolute, unless others have any different thoughts.

Dockerfile
kedark3: Ok, then I think that should be `COPY . /root/broker`

Dockerfile
kedark3: Thanks, that'd be great.

Dockerfile
kedark3: image

Dockerfile
JacobCallahan: you asked nicely  you got it :)

Dockerfile
JacobCallahan: there's no difference between the two in the context of the Dockerfile, since the build directory is the base directory.

Dockerfile
mshriver, line 6: This is redundant with the use of `WORKDIR` - it will create the directory for you

Dockerfile
mshriver: Moving this after WORKDIR will allow for using `COPY . .`, though @kedark3 requested more explicit absolute paths above so he might argue against this use.

Issue #17, opened 2020-06-18
"IndexError: pop from empty list" when checking out deploy-sat-lite workflow


mirekdlugosz:
```
$ broker checkout --workflow deploy-sat-lite --template satellite-6.8-latest
[INFO 200618 12:59:52] Using provider AnsibleTower to checkout
Traceback (most recent call last):
  File "/home/mzalewsk/.virtualenvs/broker/bin/broker", line 11, in <module>
    load_entry_point('broker', 'console_scripts', 'broker')()
  File "/home/mzalewsk/.virtualenvs/broker/lib64/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/mzalewsk/.virtualenvs/broker/lib64/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/mzalewsk/.virtualenvs/broker/lib64/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mzalewsk/.virtualenvs/broker/lib64/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/mzalewsk/.virtualenvs/broker/lib64/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/mzalewsk/.virtualenvs/broker/lib64/python3.8/site-packages/click/decorators.py", line 21, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/mzalewsk/sources/broker/broker/commands.py", line 56, in checkout
    broker_inst.checkout()
  File "/home/mzalewsk/sources/broker/broker/broker.py", line 55, in checkout
    host = self._act(provider, method, checkout=True)
  File "/home/mzalewsk/sources/broker/broker/broker.py", line 39, in _act
    return provider_inst.construct_host(
  File "/home/mzalewsk/sources/broker/broker/providers/ansible_tower.py", line 91, in construct_host
    job_attrs = self._merge_artifacts(
  File "/home/mzalewsk/sources/broker/broker/providers/ansible_tower.py", line 75, in _merge_artifacts
    child_obj = self.v2.jobs.get(id=child_id).results.pop()
IndexError: pop from empty list
```

[...]

PR #19, opened 2020-06-23
Don't throw exception when child job object can't be accessed

mirekdlugosz:
See #17 for discussion and details.
This way we might miss some "artifacts", but since I don't know what they are, I am not concerned about them.

```
$ broker checkout --workflow deploy-sat-lite --template satellite-6.8-latest
[INFO 200623 19:16:45] Using provider AnsibleTower to checkout
[INFO 200623 19:32:41] Host: dhcp-2-66.REDACTED.redhat.com
```


JacobCallahan:
This should almost always have artifacts, so I'm looking into the underlying cause now. I will consider PR after.
Thanks!

PR #21, opened 2020-06-25
Update container image

kedark3:
The current Docker image was being built as root. It is a security issue as well as not works well with OpenShift Pods. I am proposing to use `/home/broker` dir and make it `0777`. Also use `USER 1001` - an arbitrary user the container will use by default(this is just better way to build images based on my knowledge) - OpenShift will do random Username each time is creates container, so opening permissions to /home/broker will help us tackle issues about permissions.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,11 +4,13 @@ MAINTAINER https://github.com/SatelliteQE
 RUN dnf -y install make cmake gcc-c++ zlib-devel \
            openssl-devel git python3-pip python3-devel \
            && dnf clean all
-WORKDIR /root/broker
-COPY . /root/broker/
+
+WORKDIR /home/broker
+COPY . /home/broker/
 RUN pip install .
 RUN cp settings.yaml.example settings.yaml
-
+RUN chmod -Rv 0777 /home/broker
+USER 1001
 
 ENTRYPOINT ["broker"]
 CMD ["--help"]


Dockerfile
```diff
@@ -4,11 +4,13 @@ MAINTAINER https://github.com/SatelliteQE
...
-WORKDIR /root/broker
-COPY . /root/broker/
+
+WORKDIR /home/broker
+COPY . /home/broker/
 RUN pip install .
 RUN cp settings.yaml.example settings.yaml
-
+RUN chmod -Rv 0777 /home/broker
+USER 1001
```
JacobCallahan, line 13: will the user always be 1001 in openshift?

kedark3: Nope, as I was telling Justin in chat, this is just a good practice from the security POV to have containers run as non-root user. OpenShift will run as some random user, but if you run the container locally, and drop into shell, you will see it runs as 1001. I strongly recommend keeping image non-root by default.
OpenShift will get necessary perms as no matter what user we run, broker dir is given `0777`.

JacobCallahan:
Currently seeing permission issues in podman
```
podman run --rm -v $(pwd):/home/broker:Z quay.io/kkulkarn/broker:update_container_image inventory --sync AnsibleTower
[INFO 200626 14:15:43] Pulling remote inventory from AnsibleTower
[WARNING 200626 14:15:47] Unable to determine workflow for None
[INFO 200626 14:15:49] Adding new hosts: ...
Traceback (most recent call last):
  File "/usr/local/bin/broker", line 11, in <module>
    load_entry_point('broker==0.0.9', 'console_scripts', 'broker')()
  File "/usr/local/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/broker/commands.py", line 109, in inventory
    VMBroker.sync_inventory(provider=sync)
  File "/usr/local/lib/python3.7/site-packages/broker/broker.py", line 131, in sync_inventory
    helpers.update_inventory(add=new_hosts)
  File "/usr/local/lib/python3.7/site-packages/broker/helpers.py", line 111, in update_inventory
    inventory_file.unlink()
  File "/usr/lib64/python3.7/pathlib.py", line 1304, in unlink
    self._accessor.unlink(self)
PermissionError: [Errno 13] Permission denied: 'inventory.yaml'
```

kedark3:
```podman run -u 0 --rm -v $(pwd):/home/broker:Z quay.io/kkulkarn/broker:update_container_image inventory --sync AnsibleTower```

You can do `-u 0` when you want to run this container as an executable for the broker. It is not ideal either, but its better than running it as root by default. Also, do we really envision users wanting to mount the volume and run the broker as we are suggesting vs just pip installing broker? 

Another option could be ```podman run --rm -it  -v $(pwd)/settings.yaml:/home/broker/settings.yaml:Z  --entrypoint /bin/bash broker``` Where you can just map settings and it can read correct settings in the container.
I suspect you want to be able to write the data to inventory and log file from container and make it persist?

JacobCallahan:
```podman run -u 0 --rm -v $(pwd):/home/broker:Z quay.io/kkulkarn/broker:update_container_image inventory --sync AnsibleTower``` had the same result of not being able to read the settings file.

I do expect and want to support users using broker in this manner. Persistent inventory is key to container-ized broker

kedark3:
@JacobCallahan I think it is a limitation of Podman/Docker.  I couldn't find a way to mount as a non-root user. Anything we mount gets mounted as root. So if you insist on your use case, I think we have to let go the security and remove the line `USER 1001`. @mshriver you have any thoughts?

JacobCallahan:
@kedark3 how about introducing this as an Openshift Dockerfile? That could use the :) insecure :) root user while the openshift one can be more locked down.

kedark3:
@JacobCallahan I'd be happy to do so. Should I update the PR? Just wanted to see if @mshriver and you both are on board with that before I do that.

JacobCallahan:
@kedark3 that works for me! Also, you might want to add in the BROKER_DIRECTORY as an environment variable just to be sure :)

kedark3:
Closing it for now as some of these things were added to internal gitlab repo.

PR #23, opened 2020-06-29
Updated broker to be compatible with dynaconf 3.0.0

JacobCallahan:
+ Added the ability to specify a BROKER_DIRECTORY envrionment variable
+ Changed settings.yaml to broker_settings.yaml


JacobCallahan:
This PR should not be merged until Dynaconf 3.0 is released and the setup.py file reverts back to its usual dynaconf requirement. The current state is for testing purposes only.

README.md
```diff
@@ -9,9 +9,13 @@ Broker is a tool designed to provide a common interface between one or many serv
...
 pip install .
-cp setting.yaml.example settings.yaml
+cp broker_settings.yaml.example broker_settings.yaml
 ```
-Then edit the settings.yaml file
+Then edit the broker_settings.yaml file
+
+Broker can also be ran outside of its base directory. In order to do so, specify the directory broker's files are in with the
+`BROKER_DIRECTORY` envronment variable.
+```BROKER_DIRECTORY=/home/jake/Programming/broker/ broker inventory```
```
kedark3, line 18: @mshriver mentioned that we can also individually override username/password in settings.yml file using env vars. Can you add an example of that @JacobCallahan, please?

JacobCallahan: @rochacbruno can correct me if i'm wrong, but this should follow the pattern of `DYNACONF_ANSIBLETOWER_USERNAME=myusername    DYNACONF_ANSIBLETOWER_PASSWORD=mypassword`
`DYNACONF_HOST_USERNAME=myusername   DYNACONF_HOST_PASSWORD=mypassword`

mshriver: @kedark3 this is completely controlled by dynaconf, its not a robottelo function.

broker/helpers.py
```diff
@@ -84,6 +85,10 @@ def load_inventory():
     :return: list of dictionaries
     """
     inventory_file = Path(settings.INVENTORY_FILE)
+    if "BROKER_DIRECTORY" in os.environ:
```
mshriver, line 88: I'm skeptical  of this pattern, trying to resolve the envvar every time we want to access one of the files.

Can we create a singleton within the settings module that specifies the inventory file - doing the envvar processing in one place - which is then imported for use within the modules that reference it?

JacobCallahan: that's an option as well. i initially kept it as this, since it is essentially an O(1) dictionary lookup, but this could be handled in the new settings file

mshriver: Not too concerned about the execution time of the lookup - just the pattern of having to repeat this 4 line block every time we want to reference the inventory file.

setup.py
```diff
@@ -4,6 +4,11 @@
     "@11.2.0#egg=awxkit&subdirectory=awxkit"
 )
 
+DYNACONF = (
+    "dynaconf @ git+https://github.com/rochacbruno/dynaconf.git"
```
mshriver, line 8: The real 3.0 is out now! :tada:

README.md
rochacbruno: @JacobCallahan if you set `envvar_prefix="BROKER"` then you van have `BROKER_FOO=bar` to set the value of `settings.foo` instead of `DYNACONF`.

PR #24, opened 2020-06-30
Fix broker skipping every other host during checkin --all

JacobCallahan:
This allows broker to truly checkin all hosts at once

### DIFF
--- a/broker/broker.py
+++ b/broker/broker.py
@@ -89,7 +89,14 @@ class VMBroker:
             self._act(provider, "nick_help", checkout=False)
 
     def checkin(self, host=None):
-        """checkin one or more VMs"""
+        """checkin one or more VMs
+
+        :param host: can be one of:
+            None - Will use the contents of self._hosts
+            A single host object
+            A list of host objects
+            A dictionary mapping host types to one or more host objects
+        """
         if host is None:
             host = self._hosts
         logger.debug(host)
@@ -97,7 +104,8 @@ class VMBroker:
             for _host in host.values():
                 self.checkin(_host)
         elif isinstance(host, list):
-            for _host in host:
+            # reversing over a copy of the list to avoid skipping
+            for _host in host[::-1]:
                 self.checkin(_host)
         elif host:
             logger.info(f"Checking in {host.hostname}")


JacobCallahan:
I've tested locally, and it worked :)
@bherring could apply this change locally and test against his inventory

broker/broker.py
```diff
@@ -91,7 +91,7 @@ def nick_help(self):
     def checkin(self, host=None):
         """checkin one or more VMs"""
         if host is None:
-            host = self._hosts
+            host = self._hosts[:]
```
mshriver, line 94: Not to blow your scope up here, but I'm assuming the shallow copy that happens because of this slice is preventing odd behavior within the for loops on L97/100.

Its really black-box implicit, which I'm skeptical of. Immediate question when reading this code outside the context of this PR would be  - why is this being sliced?

Consider changing up the pattern of the method signature here, which may help clean up some of the logic paths.

Take `*hosts` in the function signature, and assign `hosts_to_remove = hosts.values() if hosts else copy(self._hosts)`, and update the docblock to more clearly explain what data types host should be.

Then the copy is explicit, and you don't have to have a check for the data type, and separate `for` loops both calling `checkin`.

JacobCallahan: so this method is built to allow 4 separate patterns of checkins. 
1. No host is specified, so it checks in everything broker has in its self._hosts variable, which could be any of the 3 following.
2. A single host, which is a host object.
3. A list of hosts all of the same host type.
4. A dictionary mapping host type names to one or more host objects.

Really the most effective way to handle all cases isn't the change I made here (realizing after-the-fact) but what I'm about to force push up in a new change!

mshriver: Sounds like we need a bit more consistency in `self._hosts` even - looking forward to the update.

mshriver: Taking `*hosts` in the signature at least eliminates #2 in your list, since you won't get a single host instance assigned to the kwarg

broker/broker.py
```diff
@@ -97,7 +97,7 @@ def checkin(self, host=None):
             for _host in host.values():
                 self.checkin(_host)
         elif isinstance(host, list):
-            for _host in host:
+            for _host in host[::-1]:
```
mshriver, line 100: @JacobCallahan and I had a long talk about the hole we might be digging here, concerning whether `self._hosts` is ever a dictionary, or if a dictionary would ever be passed to the `checkin` call.

Please add an inline comment here for why the list is reversed, since the reason is quite opaque and has to do with the C implementation of list iteration.

JacobCallahan: I'll add a comment there and also beef up the "docstring" for this method

bherrin3:
# Result
* Works as expected
* Linear checkin of multiple resource (10 in this instance) takes a while (5mins)

## Testing Methodology

* Created a set of hosts in different ways, not necessarily from Broker
* Sync'd the AT inventory sources to ensure the Provider/AT inventory matched
* Ran `broker inventory --sync AnsibleTower` to ensure the sync matched all inventories
* Checked in all VMs with `broker checkin --all` and verified completion
* Sync'd the AT inventory sources to ensure the Provider/AT inventory matched
* Ran `broker inventory --sync AnsibleTower` to ensure the sync matched all inventories

## Results

```
(jake) ➜  broker git:(master) ✗ broker inventory --sync AnsibleTower
[INFO 200630 16:25:41] Pulling remote inventory from AnsibleTower
[INFO 200630 16:25:43] Adding new hosts: dhcp-3-43.vms.sat.rdu2.redhat.com
[INFO 200630 16:25:43] Pulling local inventory
[INFO 200630 16:25:43] 0: dhcp-3-23.vms.sat.rdu2.redhat.com
[INFO 200630 16:25:43] 1: dhcp-2-17.vms.sat.rdu2.redhat.com
[INFO 200630 16:25:43] 2: dhcp-2-62.vms.sat.rdu2.redhat.com
[INFO 200630 16:25:43] 3: dhcp-2-26.vms.sat.rdu2.redhat.com
[INFO 200630 16:25:43] 4: dhcp-2-32.vms.sat.rdu2.redhat.com
[INFO 200630 16:25:43] 5: dhcp-2-122.vms.sat.rdu2.redhat.com
[INFO 200630 16:25:43] 6: dhcp-3-79.vms.sat.rdu2.redhat.com
[INFO 200630 16:25:43] 7: dhcp-2-106.vms.sat.rdu2.redhat.com
[INFO 200630 16:25:43] 8: dhcp-3-78.vms.sat.rdu2.redhat.com
[INFO 200630 16:25:43] 9: dhcp-2-254.vms.sat.rdu2.redhat.com
[INFO 200630 16:25:43] 10: dhcp-3-43.vms.sat.rdu2.redhat.com
(jake) ➜  broker git:(master) ✗ broker checkin --all
[INFO 200630 16:27:55] Checking in dhcp-3-43.vms.sat.rdu2.redhat.com
[INFO 200630 16:28:26] Checking in dhcp-2-254.vms.sat.rdu2.redhat.com
[INFO 200630 16:28:58] Checking in dhcp-3-78.vms.sat.rdu2.redhat.com
[INFO 200630 16:29:24] Checking in dhcp-2-106.vms.sat.rdu2.redhat.com
[INFO 200630 16:29:57] Checking in dhcp-3-79.vms.sat.rdu2.redhat.com
[INFO 200630 16:30:29] Checking in dhcp-2-122.vms.sat.rdu2.redhat.com
[INFO 200630 16:31:01] Checking in dhcp-2-32.vms.sat.rdu2.redhat.com
[INFO 200630 16:31:27] Checking in dhcp-2-26.vms.sat.rdu2.redhat.com
[INFO 200630 16:31:59] Checking in dhcp-2-62.vms.sat.rdu2.redhat.com
[INFO 200630 16:32:26] Checking in dhcp-2-17.vms.sat.rdu2.redhat.com
[INFO 200630 16:32:57] Checking in dhcp-3-23.vms.sat.rdu2.redhat.com
(jake) ➜  broker git:(master) ✗ broker inventory --sync AnsibleTower
[INFO 200630 16:36:23] Pulling remote inventory from AnsibleTower
[INFO 200630 16:36:24] No new hosts found
[INFO 200630 16:36:24] Pulling local inventory
(jake) ➜  broker git:(master) ✗ 
```

PR #25, opened 2020-07-02
Add background mode to Broker cli

JacobCallahan:
This change introduces the ability for broker to spin up another broker
subprocess and exit.
This new broker subprocess will also have a new log-level of "silent".
A "silent" log level has a level of INFO, but only logs to broker's log file.

### DIFF
--- a/broker/commands.py
+++ b/broker/commands.py
@@ -1,3 +1,4 @@
+import subprocess
 from pathlib import Path
 import click
 import logging
@@ -8,16 +9,28 @@ from broker import helpers
 from broker.hosts import Host
 
 
+def _args_to_cmd_str(args_dict):
+    cmd_str = ""
+    for key, value in args_dict.items():
+        cmd_str = f"{cmd_str}{f'--{key}' if not key.startswith('--') else key} {value} "
+    return cmd_str
+
+
 def update_log_level(ctx, param, value):
-    if getattr(logging, value.upper()) is not logger.getEffectiveLevel():
-        b_log.setup_logzero(level=value)
-        click.echo(f"Log level changed to [{value}]")
+    silent = False
+    if value == "silent":
+        silent = True
+        value = "info"
+    if getattr(logging, value.upper()) is not logger.getEffectiveLevel() or silent:
+        b_log.setup_logzero(level=value, silent=silent)
+        if not silent:
+            click.echo(f"Log level changed to [{value}]")
 
 
 @click.group()
 @click.option(
     "--log-level",
-    type=click.Choice(["info", "warning", "error", "critical", "debug"]),
+    type=click.Choice(["info", "warning", "error", "critical", "debug", "silent"]),
     default="info",
     callback=update_log_level,
     is_eager=True,
@@ -30,16 +43,19 @@ def cli():
 @cli.command(
     context_settings={"allow_extra_args": True, "ignore_unknown_options": True}
 )
+@click.option("-b", "--background", is_flag=True, help="Run checkout in the background")
 @click.option("--workflow", type=str)
 @click.option("--nick", type=str, help="Use a nickname defined in your settings")
 @click.pass_context
-def checkout(ctx, workflow, nick):
+def checkout(ctx, background, workflow, nick):
     """Checkout or "create" a Virtual Machine broker instance
     COMMAND: broker checkout --workflow "workflow-name" --workflow-arg1 something
     or
     COMMAND: broker checkout --nick "nickname"
 
     :param ctx: clicks context object
+
+    :param background: run a new broker subprocess to carry out command
 
     :param workflow: workflow template stored in Ansible Tower, passed in as a string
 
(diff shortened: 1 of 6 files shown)


broker/settings.py
```diff
@@ -1,13 +1,16 @@
 import os
 from pathlib import Path
 from dynaconf import Dynaconf
+from logzero import logger
 
 settings_file = "broker_settings.yaml"
 BROKER_DIRECTORY = Path()
+logger.info(BROKER_DIRECTORY.absolute())
```
omkarkhatavkar, line 8: is these line causing
```
 (env) /home/okhatavk/Satellite/broker (test_pr_!) $ broker --help 
[I 200706 20:25:56 settings:8] /home/okhatavk/Satellite/broker
[I 200706 20:25:56 settings:15] /home/okhatavk/Satellite/broker

```

broker/settings.py
```diff
@@ -1,13 +1,16 @@
...
 
 if "BROKER_DIRECTORY" in os.environ:
     envar_location = Path(os.environ["BROKER_DIRECTORY"])
     if envar_location.is_dir():
         BROKER_DIRECTORY = envar_location
 
+logger.info(BROKER_DIRECTORY.absolute())
```
omkarkhatavkar, line 15: is this needed ?

broker/settings.py
JacobCallahan: ahhh nice catch, had this from my testing. could remove this or make it debug. probably the latter

broker/settings.py
JacobCallahan: see comment above :)

broker/commands.py
```diff
@@ -81,26 +108,39 @@ def nick_help(ctx, workflow, provider):
 
 @cli.command()
 @click.argument("vm", type=str, nargs=-1)
+@click.option("-d", "--daemon", is_flag=True, help="Run checkout in the background")
```
omkarkhatavkar, line 111:
```suggestion
@click.option("-d", "--daemon", is_flag=True, help="Run checkin in the background")
```

broker/commands.py
```diff
@@ -117,10 +157,12 @@ def inventory(details, sync):
         else:
             logger.info(f"{num}: {host['hostname'] or host['name']}")
 
+
 @cli.command()
 @click.argument("vm", type=str, nargs=-1)
+@click.option("-d", "--daemon", is_flag=True, help="Run checkout in the background")
```
omkarkhatavkar, line 163:
```suggestion
@click.option("-d", "--daemon", is_flag=True, help="Run duplicate in the background")
```

broker/commands.py
```diff
@@ -129,29 +171,44 @@ def duplicate(vm, all_):
...
+                else:
+                    logger.warning(
+                        f"Unable to duplicate {host['hostname']}, no _broker_args found"
+                    )
 
 
 @cli.command(
     context_settings={"allow_extra_args": True, "ignore_unknown_options": True}
 )
+@click.option("-d", "--daemon", is_flag=True, help="Run checkout in the background")
```
omkarkhatavkar, line 199:
```suggestion
@click.option("-d", "--daemon", is_flag=True, help="Run execute in the background")
```

broker/commands.py
```diff
@@ -52,15 +68,28 @@ def checkout(ctx, workflow, nick):
         broker_args["workflow"] = workflow
     # if additional arguments were passed, include them in the broker args
     broker_args.update(dict(zip(ctx.args[::2], ctx.args[1::2])))
-    broker_inst = VMBroker(**broker_args)
-    broker_inst.checkout()
+    if background:
+        logger.info("Running checkout in the background.")
+        subprocess.Popen(
+            f"broker --log-level silent checkout {_args_to_cmd_str(broker_args)}",
```
mshriver, line 74: There is some platform dependency in not passing this as `args` list, but as a single string.

broker/commands.py
```diff
@@ -52,15 +68,28 @@ def checkout(ctx, workflow, nick):
...
+            shell=True,
+        )
+    else:
```
mshriver, line 77: Something to consider - we can fork and exit the main process, instead of using `Popen`. This also means that we won't have to convert the kwargs here back to CLI args ... for them to be re-read and converted to kwargs again.

broker/commands.py
```diff
@@ -52,15 +68,28 @@ def checkout(ctx, workflow, nick):
         broker_args["workflow"] = workflow
     # if additional arguments were passed, include them in the broker args
     broker_args.update(dict(zip(ctx.args[::2], ctx.args[1::2])))
-    broker_inst = VMBroker(**broker_args)
-    broker_inst.checkout()
+    if background:
+        logger.info("Running checkout in the background.")
+        subprocess.Popen(
```
mshriver, line 73: Please label this Popen instance and log (debug level) the `Popen.pid` property so users can positively identify child processes.

JacobCallahan: ahh missed one!

JacobCallahan: ahhh this was discussed and demonstrated, but not truly added to this pr. i can update

JacobCallahan:
Closing in favor of #28

PR #26, opened 2020-07-07
Wrap IndexError with log and exit condition

mshriver:
Handle the failure to find a workflow by name a little bit better, and log more information when executing workflows.

~~@JacobCallahan thoughts on the sys.exit call here? Would you rather a different pattern?~~

### Error Handling
This used to result in an IndexError and traceback dump, captured and logged without the trace.

```
setup@localhost:~/repos/broker (handle-workflow-indexerror*>) % broker execute  --workflow does-not-exist
[INFO 200707 14:18:28] Using provider AnsibleTower for execution
[ERROR 200707 14:18:29] Workflow not found by name: does-not-exist
[INFO 200707 14:18:29] None
```

### New Logging
Add logging for the workflow template URL and the job URL, so the user can more easily identify and track their request.

```
setup@localhost:~/repos/broker (handle-workflow-indexerror*>) % broker execute  --workflow workflow-broker-test
[INFO 200707 14:14:29] Using provider AnsibleTower for execution
[INFO 200707 14:14:31] Found workflow by name [workflow-broker-test] with description: I made this part up
[INFO 200707 14:14:31] Launching workflow template: https://<ansible-tower-url>/api/v2/workflow_job_templates/53/
[INFO 200707 14:14:31] Waiting for job: https://<ansible-tower-url>/api/v2/workflow_jobs/6455/
```

### DIFF
--- a/broker/providers/ansible_tower.py
+++ b/broker/providers/ansible_tower.py
@@ -148,8 +148,15 @@ class AnsibleTower(Provider):
         if "artifacts" in kwargs:
             kwargs.pop("artifacts")
             artifacts = True
-        wfjt = self.v2.workflow_job_templates.get(name=workflow).results.pop()
+        wfjts = self.v2.workflow_job_templates.get(name=workflow).results
+        if wfjts:
+            wfjt = wfjts.pop()
+        else:
+            logger.error(f'Workflow not found by name: {workflow}')
+            return
+        logger.debug(f'Launching workflow template: {AT_URL}{wfjt.url}')
         job = wfjt.launch(payload={"extra_vars": str(kwargs).replace("--", "")})
+        logger.info(f'Waiting for job: {AT_URL}{job.url}')
         job.wait_until_completed(timeout=1800)
         if not job.status == "successful":
             logger.error(
--- a/tests/data/ansible_tower/fake_workflows.json
+++ b/tests/data/ansible_tower/fake_workflows.json
@@ -1,0 +1,8 @@
+[
+  {
+    "id": 34,
+    "type": "workflow_job_template",
+    "url": "/api/v2/workflow_job_templates/34/",
+    "name": "deploy-base-rhel"
+  }
+]
--- a/tests/providers/test_ansible_tower.py
+++ b/tests/providers/test_ansible_tower.py
@@ -24,13 +24,27 @@ class AwxkitApiStub(MockStub):
         if "job_id" in kwargs:
             # we're a job, so load in job information
             super().__init__(self._load_job(kwargs.pop("job_id")))
+        elif "name" in kwargs:
+            # workflow job template lookup
+            super().__init__(self._load_workflow(kwargs.pop("name")))
+        else:
+            super().__init__()
 
-    def _load_job(self, job_id):
+    @staticmethod
+    def _load_job(job_id):
         with open("tests/data/ansible_tower/fake_jobs.json") as job_file:
             job_data = json.load(job_file)
         for job in job_data:
             if job["id"] == job_id:
                 return job
+
+    @staticmethod
+    def _load_workflow(workflow_name):
+        with open("tests/data/ansible_tower/fake_workflows.json") as workflow_file:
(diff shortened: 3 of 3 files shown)


broker/providers/ansible_tower.py
```diff
@@ -148,7 +149,11 @@ def exec_workflow(self, **kwargs):
         if "artifacts" in kwargs:
             kwargs.pop("artifacts")
             artifacts = True
-        wfjt = self.v2.workflow_job_templates.get(name=workflow).results.pop()
+        try:
+            wfjt = self.v2.workflow_job_templates.get(name=workflow).results.pop()
+        except IndexError:
+            logger.exception(f'Workflow not found by name: {workflow}')
+            sys.exit(1)
```
JacobCallahan, line 156: how about this?
```
        try:
            wfjt = self.v2.workflow_job_templates.get(name=workflow).results.pop()
        except IndexError:
            logger.warning(f'Workflow not found by name: {workflow}')
            return
```

JacobCallahan:
Screenshot from 2020-07-02 11-01-19
example of the result of the change i've suggested

JacobCallahan:
Screenshot from 2020-07-07 11-58-46
example of the change i've suggested

broker/providers/ansible_tower.py
mshriver: I guess the exception context of the log message is useful when working on broker, but not as necessary for regular use.  Seeing that exception short trace pollutes the console output otherwise, even though the initial message is formatted.

broker/providers/ansible_tower.py
```diff
@@ -148,8 +149,16 @@ def exec_workflow(self, **kwargs):
         if "artifacts" in kwargs:
             kwargs.pop("artifacts")
             artifacts = True
-        wfjt = self.v2.workflow_job_templates.get(name=workflow).results.pop()
+        wfjts = self.v2.workflow_job_templates.get(name=workflow).results
+        if wfjts:
+            wfjt = wfjts.pop()
+        else:
+            logger.error(f'Workflow not found by name: {workflow}')
+            return
```
JacobCallahan, line 157: I like this version!

broker/providers/ansible_tower.py
```diff
@@ -148,8 +149,16 @@ def exec_workflow(self, **kwargs):
...
+        logger.info(f'Found workflow by name [{workflow}] with description: {wfjt.description}')
+        logger.info(f'Launching workflow template: {AT_URL}{wfjt.url}')
         job = wfjt.launch(payload={"extra_vars": str(kwargs).replace("--", "")})
+        logger.info(f'Waiting for job: {AT_URL}{job.url}')
```
JacobCallahan, line 161: I'll be honest, I'm not a fan of this much logging.

mshriver: I was on the fence about including these as info level, but I think there's value to this verbosity since broker is just the interface to tools outside of its own domain.

If a user wants to observe the job or inspect the workflow template that they're calling, they currently have no easy mechanism to link from what they're doing on the CLI to what is happening on the provider.

Even more so, these external entities that we're resolving and calling could change completely outside of broker's scope. A workflow could be renamed, and the call that you made yesterday calling workflow template ID 1 could today end up resolving workflow template ID 2.

With multiple people calling the same workflow template, it could be time consuming, if not difficult, for them to look back through the jobs for that template and match one to the request made from their CLI.

Without some explicit logging about what external entities we're resolving, the user is left in the dark about what happened on previous checkouts, and what's happening in the current checkout.

broker/providers/ansible_tower.py
mshriver: unfortunately the MockStub class used for awx unit testing doesn't handle this and recurses or otherwise can't evaluate. Looking at updating the class to support the existing unit tests with this pattern.

mshriver: Think I've found an update to the AwxkitApiStub that supports this.

PR #28, opened 2020-07-10
[Alternate]Add background mode to Broker cli

JacobCallahan:
This change introduces the ability for broker to spin up another broker
subprocess and exit.
This new broker subprocess will also have a new log-level of "silent".
A "silent" log level has a level of INFO, but only logs to broker's log file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ You may use either the local id (```broker inventory```), the hostname, or "all"
 broker checkin my.host.fqdn.com
 broker checkin 0
 broker checkin 1 3 my.host.fqdn.com
-broker checkin all
+broker checkin --all
 ```
 
 **Creating nicks**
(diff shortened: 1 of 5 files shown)


JacobCallahan:
This PR serves as an alternate implementation to #25 
The major difference being a change from subprocess.Popen to os.fork

README.md
```diff
@@ -54,7 +54,7 @@ You may use either the local id (```broker inventory```), the hostname, or "all"
 broker checkin my.host.fqdn.com
 broker checkin 0
 broker checkin 1 3 my.host.fqdn.com
-broker checkin all
+broker checkin --all
```
mshriver, line 57: sneaky

broker/commands.py
```diff
@@ -1,23 +1,36 @@
...
+        if not silent:
+            click.echo(f"Log level changed to [{value}]")
+
+
+def fork_broker():
+    pid = os.fork()
+    if pid:
+        logger.info(f"Running broker in the background with pid: {pid}")
+        sys.exit(0)
+    update_log_level(None, None, "silent")
```
mshriver, line 27: Commenting here, but the change would have to go into `setup_logzero`

Not a blocking request.

 I think it would be ideal for the log message prefix to include the PID, at least when running in background mode.

README.md
JacobCallahan: that was noticed by none other than my bherring himself!

broker/commands.py
JacobCallahan: interesting proposition

PR #29, opened 2020-07-27
settings_broker.yaml to settings.yaml

tstrych:
Is that `broker_settings` ? 
In my workflow only `settings.yaml` works.


tstrych:
My bad,  there is already broker 0.11, and forgot to update my venv. In broker 0.11 there is `broker_settings.yaml`

PR #31, opened 2020-07-31
Added ability to extend vm lease time

JacobCallahan:
Changed --artifacts to now accept merge or last
Misc small changes

**Note:** This adds a new field in the ansible tower settings for `extend_workflow`

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,19 @@ Broker maintains a local inventory of the VMs you've checked out. You can see th
 ```
 broker inventory
 ```
+To sync your inventory from a supported provider, use the `--sync` option.
+```
+broker inventory --sync AnsibleTower
+```
+
+**Extending your VM lease time***
+Providers supporting extending a VM's lease time make that functionality available through the `extend` subcommand.
+```
+broker extend 0
+broker extend hostname
+broker extend vmname
+broker extend --all
+```
 
 **Checking in VMs**
 You can also return a VM to its provider with the ```checkin``` command.
@@ -75,6 +88,7 @@ If a provider action doesn't result in a host creation/removal, Broker allows yo
 broker execute --help
 broker execute --workflow my-awesome-workflow --additional-arg True
 broker execute -o raw --workflow my-awesome-workflow --additional-arg True
+broker execute -o raw --workflow my-awesome-workflow --additional-arg True --artifacts last
 ```
 
 **Run Broker in the background**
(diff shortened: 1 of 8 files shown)


broker/commands.py
```diff
@@ -193,8 +211,8 @@ def duplicate(vm, background, all_):
 )
 @click.option(
     "--artifacts",
-    is_flag=True,
-    help="AnsibleTower: return all artifacts associated with the execution.",
+    type=click.Choice(["merge", "last"]),
```
mshriver, line 214: Duplicates the default in the function signature, but I think its worth setting a default here explicitly as well.

JacobCallahan: if you set a duplicate, then you're unable to get the raw job information, as artifacts will always be set. (i tried this)

broker/commands.py
```diff
@@ -129,15 +129,31 @@ def checkin(vm, background, all_):
...
+@click.option(
+    "--extend", type=str, multiple=True, help="Extend the lease of one or more hosts"
+)
 @click.option(
     "--sync",
     type=str,
     help="Class-style name of a supported broker provider. (AnsibleTower)",
 )
-def inventory(details, sync):
+def inventory(details, extend, sync):
```
mshriver, line 140: Consider turning `inventory` into a command group, with `extend` and `sync` as subcommands, instead of having to check each of those inventory 'modes' within the actual inventory command.

That or make it a command separate from `inventory`, like `duplicate`. I would think `duplicate` and `extend` would operate with a nearly identical argument structure, operating on a given list of inventory items.

The pattern of implementation of `extend` is nearly identical to `duplicate`, without the support for backgrounding or passing 'all'.  I think plenty of users are going to want to extend all of their checkouts.

broker/providers/ansible_tower.py
```diff
@@ -16,6 +16,7 @@
 UNAME = settings.ANSIBLETOWER.username
 PWORD = settings.ANSIBLETOWER.password
 RELEASE_WORKFLOW = settings.ANSIBLETOWER.release_workflow
+EXTEND_WORFLOW = settings.ANSIBLETOWER.extend_workflow
```
mshriver, line 19: Missing a `K` here

README.md
```diff
@@ -46,6 +46,17 @@ Broker maintains a local inventory of the VMs you've checked out. You can see th
 ```
 broker inventory
 ```
+To sync your inventory from a supported provider, use the `--sync` option.
+```
+broker inventory --sync AnsibleTower
+```
+Providers supporting extending a VM's lease time make that functionality available through the `--extend` option.
+```
+broker inventory --extend 0
```
mshriver, line 55: This block needs to be updated to reflect the changed command structure.

PR #32, opened 2020-08-03
Bring your own host classes

JacobCallahan:
This change adds the ability to override or supply your own
host classes via the VMBroker class.
VMBroker now looks for a kwarg called "host_classes" which should be
a dictionary mapping class name to the class object.

### DIFF
--- a/broker/broker.py
+++ b/broker/broker.py
@@ -13,16 +13,17 @@ PROVIDER_ACTIONS = {
     "test_action": (TestProvider, "test_action"),
 }
 
-HOST_CLASSES = {"host": Host}
-
 
 class VMBroker:
     def __init__(self, **kwargs):
         self._hosts = kwargs.pop("hosts", [])
+        self.host_classes = {"host": Host}
         # if a nick was specified, pull in the resolved arguments
         if "nick" in kwargs:
             nick = kwargs.pop("nick")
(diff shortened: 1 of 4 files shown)


broker/broker.py
```diff
@@ -23,6 +23,10 @@ def __init__(self, **kwargs):
         if "nick" in kwargs:
             nick = kwargs.pop("nick")
             kwargs = helpers.merge_dicts(kwargs, helpers.resolve_nick(nick))
+        if "host_classes" in kwargs:
+            classes = kwargs.pop("host_classes")
+            global HOST_CLASSES
```
mshriver, line 28: What's the purpose of this being global, instead of an instance attribute?

JacobCallahan: Global is from a previous design decision. But this could be instance-based

broker/providers/ansible_tower.py
```diff
@@ -72,16 +72,23 @@ def _merge_artifacts(self, at_object, strategy="merge", artifacts=None):
             elif strategy == "last":
                 artifacts = at_object.artifacts
         if "workflow_nodes" in at_object.related:
-            children = at_object.get_related("workflow_nodes").results
+            children = at_object.get_related("workflow_nodes").results.sort(
```
mshriver, line 75: The `sort()` call here operates on the list - but returns None, so the iteration on children fails.

PR #35, opened 2020-08-07
adding the property to see the vm host expire time

omkarkhatavkar:
### PR Details
Adding expire time property for details command, this will require sync and needs cleaning of the inventory file. 

### Result 

```
(env) [root@okhatavk broker]# broker inventory --details 
[INFO 200807 20:08:12] Pulling local inventory
[INFO 200807 20:08:12] 0: hostname_1.redhat.com, Details: _broker_args:
      host_type: satellite
      provider: RHEV
      rhel_version: '7.7'
      template: satellite-6.8.0-10.0
      workflow: deploy-sat-lite
    _broker_provider: AnsibleTower
    expire_time: '2020-08-10 15:31:28'
    hostname: hostname_1.redhat.com
    name: okhatavk-satellite-6.8.0-10.0-1596793936
    type: host
    
```

### DIFF
--- a/broker/providers/ansible_tower.py
+++ b/broker/providers/ansible_tower.py
@@ -1,5 +1,6 @@
 import inspect
 import json
+import datetime
 from broker.settings import settings
 from logzero import logger
 
@@ -91,11 +92,16 @@ class AnsibleTower(Provider):
                         )
         return artifacts
 
+    def _get_expire_date(self, host_id):
+        time_stamp = self.v2.hosts.get(id=host_id).results[0].related.ansible_facts.get().expire_date
+        return str(datetime.datetime.fromtimestamp(int(time_stamp)))
+
     def _compile_host_info(self, host):
         host_info = {
             "name": host.name,
             "type": host.type,
             "hostname": host.variables["fqdn"],
+            "expire_time": self._get_expire_date(host.id),
             "_broker_provider": "AnsibleTower",
         }
         if "last_job" in host.related:


JacobCallahan:
I'd like to also link this into an inventory --sync, but that would take additional thought on how to implement.
as far as checkouts are concerned this looks sane.

omkarkhatavkar:
@JacobCallahan Do you want me to present some result for this PR or Anything I need to add ?

JacobCallahan:
@omkarkhatavkar not specifically for this PR, but let's get @mshriver to review as well before merge. My earlier comment is something that would be nice for later

broker/providers/ansible_tower.py
```diff
@@ -91,11 +92,16 @@ def _merge_artifacts(self, at_object, strategy="merge", artifacts=None):
                         )
         return artifacts
 
+    def _get_expire_date(self, host_id):
+        time_stamp = self.v2.hosts.get(id=host_id).results[0].related.ansible_facts.get().expire_date
```
mshriver, line 96: I think the lookup of this fact needs to be safer, as its something that our team may be implementing, but not all `broker` users will implement.  A simple try/except should diaper any of the exceptions that could occur during the chained indexing and attribute references made here, and return an empty string.

Raises a bit more questions though - should 'expire_time' only be added to the host info dict when its actually set on the provider, or should we just include a key with an empty value, that's not relevant to some users.  Thoughts here @JacobCallahan?

I think the assumption that the expire_date, if set, is an epoch timestamp is also a dangerous one - again other teams may implement an expire_date as an iso8601 string, or just a date.

Again we can _try_ to parse it, and it could even be something configurable through `broker_settings`, but I'd start with just wrapping this value parsing in try/except and returning the raw value if it doesn't parse.

JacobCallahan: Good points. We can use this section to check if the value exists and fallback to not setting the attribute if it doesn't. As Mike points out, expire times may not be something that all teams implement.

omkarkhatavkar: If try catch works, we can return the string "expire time not avaible/set/not retrived or not exist" in case if the value comes empty ?  

I can return the raw value for now, but it seems be not that helpful as people will required to convert it know what is  
actual date and time. (I understand we can provide option to configure using broker setting file that seems to be more effort for small change, too much configuration in settings )

JacobCallahan: i'd prefer we didn't set the attribute if it doesn't exist. this reduces noise in the inventory

omkarkhatavkar: So for now I will try add try catch and test it and return epoch time, incase exception happans it wuld return the None

broker/providers/ansible_tower.py
```diff
@@ -91,11 +92,16 @@ def _merge_artifacts(self, at_object, strategy="merge", artifacts=None):
...
+        return str(datetime.datetime.fromtimestamp(int(time_stamp)))
+
     def _compile_host_info(self, host):
         host_info = {
             "name": host.name,
             "type": host.type,
             "hostname": host.variables["fqdn"],
+            "expire_time": self._get_expire_date(host.id),
```
JacobCallahan, line 104: you could remove this line here, then below this dict definition, get the results from _get_expire_time and if not None, add it to the dict.

broker/providers/ansible_tower.py
mshriver: @omkarkhatavkar you don't have to remove parsing of the epoch to a human readable datetime - it just has to be done safely, in case the `expire_date` fact isn't set to an epoch.

omkarkhatavkar: Okay done those changes.

mshriver (approved):

JacobCallahan (approved):
ACK, thanks!

JacobCallahan:
@omkarkhatavkar git has identified some merge conflicts, can you rebase and resolve them?

PR #37, opened 2020-08-08
Introducing multiprocessing for broker checkouts

JacobCallahan:
VMBroker now has the ability to multiprocess checkouts
The decorator is likely general to other VMBroker methods, but hasn't
been applied at this point.
Also, Broker is now v0.1.0

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,12 @@
 History
 =======
 
+0.1.0 (2020-08-08)
+==================
+
++ VMBroker now has the ability to multiprocess checkouts
++ Other miscellaneous enhancements and fixes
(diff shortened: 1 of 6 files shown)


JacobCallahan:
I'm debating about changing `--count` to some other name that won't potentially conflict with a legitimate checkout argument. maybe `broker-count`?
I'd like some options, please.

mshriver:
Unit tests?

akhil-jha:
How about `--enum` ? (enumerate)

JacobCallahan:
@akhil-jha I'm afraid that would be confusing for people familiar with python's enumerate built-in.

PR #38, opened 2020-08-10
Use token authentication for Tower

mshriver:
Support token authentication for AnsibleTower provider.

On the tower side, this requires an Application, which the tokens are registered under, and a global setting to allow token creation for OAuth users.

When using token, username and password settings are no longer necessary.

### DIFF
--- a/broker/providers/ansible_tower.py
+++ b/broker/providers/ansible_tower.py
@@ -1,5 +1,6 @@
 import inspect
 import json
+import sys
 from broker.settings import settings
 from logzero import logger
 
@@ -13,8 +14,9 @@ from broker.providers import Provider
 from broker import helpers
 
 AT_URL = settings.ANSIBLETOWER.base_url
-UNAME = settings.ANSIBLETOWER.username
-PWORD = settings.ANSIBLETOWER.password
+UNAME = settings.ANSIBLETOWER.get("username")
+PWORD = settings.ANSIBLETOWER.get("password")
+TOKEN = settings.ANSIBLETOWER.get("token")
 RELEASE_WORKFLOW = settings.ANSIBLETOWER.release_workflow
 EXTEND_WORKFLOW = settings.ANSIBLETOWER.extend_workflow
 AT_TIMEOUT = settings.ANSIBLETOWER.workflow_timeout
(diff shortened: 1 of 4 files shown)


broker/providers/ansible_tower.py
```diff
@@ -25,14 +27,28 @@ def __init__(self, **kwargs):
         self._construct_params = []
         config = kwargs.get("config", awxkit.config)
         config.base_url = AT_URL
-        config.credentials = {"default": {"username": UNAME, "password": PWORD}}
-        config.use_sessions = True
-        if "root" in kwargs:
```
mshriver, line 30: @JacobCallahan looking for your input here - what was the intent of support `root` instance being passed into this constructor?

mshriver: Nvm, found it, the mock stub behavior for unit tests.

broker/providers/ansible_tower.py
```diff
@@ -13,8 +14,9 @@
 from broker import helpers
 
 AT_URL = settings.ANSIBLETOWER.base_url
-UNAME = settings.ANSIBLETOWER.username
-PWORD = settings.ANSIBLETOWER.password
+UNAME = settings.ANSIBLETOWER.username.get("username")
```
mshriver, line 17: This is a temporary implementation of safety around reading these settings, until we've got better support for validators and defaults in DynaConf.

JacobCallahan: We may get a new dynaconf version soon, and this would end up unnecessary if handled by an upcoming pr

mshriver: I'm happy to rebase after that.

mshriver: Moved to validators - these `.get()` calls are still necessary though so this module has safe import.

broker/settings.py
```diff
@@ -29,6 +27,13 @@
     Validator("HOST_USERNAME", default="root"),
 ]
 settings = Dynaconf(settings_file=str(settings_path), validators=validators,)
+
+# Handle both modes of ansible tower authentication
+settings.validators.register(
+    ((Validator("ANSIBLETOWER.username", must_exist=True) & Validator("ANSIBLETOWER.password", must_exist=True))
+    | Validator("ANSIBLETOWER.token", must_exist=True))
+)
```
JacobCallahan, line 35: can this be combined with the above validators and included during the settings initialization?

mshriver: yep! Didn't think I'd be able to include it as a list item, but it works.

JacobCallahan (approved):
ACK

PR #39, opened 2020-08-12
Adding dynaconf validators to broker settings

JacobCallahan:
This change intends to make it more clear to users when their settings
need to be changed.
Additionally, some sane defaults can be added if missing.

### DIFF
--- a/broker/settings.py
+++ b/broker/settings.py
@@ -1,6 +1,9 @@
 import os
+import sys
 from pathlib import Path
-from dynaconf import Dynaconf
+from dynaconf import Dynaconf, Validator
+from dynaconf.validator import ValidationError
+from logzero import logger
 
 settings_file = "broker_settings.yaml"
 BROKER_DIRECTORY = Path()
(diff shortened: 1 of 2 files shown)


JacobCallahan:
Note that this change requires a new version of dynaconf. The pr is likely to change after this update is released.

JacobCallahan:
Tested changes locally and these validators are working as expected. However, this is blocked by https://github.com/rochacbruno/dynaconf/issues/391

broker/settings.py
```diff
@@ -10,4 +13,24 @@
...
         BROKER_DIRECTORY = envar_location
 
-settings = Dynaconf(settings_file=str(BROKER_DIRECTORY.joinpath("broker_settings.yaml")))
+settings_path = BROKER_DIRECTORY.joinpath("broker_settings.yaml")
+must_exist = [
+    "ANSIBLETOWER.base_url",
+    "ANSIBLETOWER.username",
+    "ANSIBLETOWER.password",
+    "NICKS",
+    "HOST_PASSWORD",
```
mshriver, line 22: Not for this PR, but I think we should consider structuring these keys in settings, like the below example. It will allow more flexibility and specification of host credentials. They could be different for different providers, or different workflows. Maybe we can tie nicks to host credential keys. A thought to chew on.

```
host:
    default:
        username:
        password:
```

broker/settings.py
```diff
@@ -10,4 +13,24 @@
...
+    Validator("ANSIBLETOWER.release_workflow", default="remove-vm"),
+    Validator("ANSIBLETOWER.extend_workflow", default="extend-vm"),
+    Validator("ANSIBLETOWER.workflow_timeout", is_type_of=int, default=3600),
+    Validator("HOST_USERNAME", default="root"),
+]
+settings = Dynaconf(settings_file=str(settings_path), validators=validators,)
+try:
+    settings.validators.validate()
+except ValidationError as err:
+    logger.error(f"Configuration error in {settings_path.absolute()}: {err}")
```
mshriver, line 35: Using logger.exception will automatically include the traceback after your error message, so you don't have to label or format in the exception.

broker/settings.py
JacobCallahan: that is certainly something we can keep in mind

broker/settings.py
JacobCallahan: let's leave this as-is for now, as i currently think the error version looks better. but we can always reconsider later

PR #40, opened 2020-08-12
Adjust for nodes without results

JacobCallahan:
This is a potentially common encounter for more complex workflow
paths. This change skips all "transition" nodes.


broker/providers/ansible_tower.py
```diff
@@ -74,6 +74,12 @@ def _merge_artifacts(self, at_object, strategy="merge", artifacts=None):
                 artifacts = at_object.artifacts
         if "workflow_nodes" in at_object.related:
             children = at_object.get_related("workflow_nodes").results
+            # filter out children with no associated job
```
mshriver, line 77: "freeloaders"

PR #49, opened 2020-09-15
Update tower logging to provide UI URLs

mshriver:
UI urls are a bit more user friendly

Use urllib.parse instead of mangling strings for URLs

Before, logged URLS had `http:/<host>/<path>`, which on console isn't parsed as a URI, and can't be directly clicked/linked. This was due to using `.replace()` to resolve any double slashes from the host ending with a slash and the path starting with a slash.

Output now looks like:
```
[INFO 200915 08:07:43] Using provider AnsibleTower to checkout
[INFO 200915 08:07:43] Using username and password authentication
[INFO 200915 08:07:45] Waiting for job: 
    API: https://<host>/api/v2/workflow_jobs/1363/
    UI: https://<host>/#/workflows/1363
```

### DIFF
--- a/broker/providers/ansible_tower.py
+++ b/broker/providers/ansible_tower.py
@@ -1,6 +1,7 @@
 import inspect
 import json
 import sys
+from urllib import parse as url_parser
 from broker.settings import settings
 from logzero import logger
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


bherrin3 (approved):

broker/providers/ansible_tower.py
```diff
@@ -210,11 +211,11 @@ def exec_workflow(self, **kwargs):
         else:
             logger.error(f"Workflow not found by name: {workflow}")
             return
-        logger.debug(
-            f"Launching workflow template: {AT_URL}{wfjt.url}".replace("//", "/")
-        )
+        logger.debug(f"Launching workflow template: {url_parser.urljoin(AT_URL, str(wfjt.url))}")
```
mshriver, line 214: When trying to pass `wfjt.url` directly, this blew up as its not actually a string, so I had to wrap it and the `job.url` reference.  Its due to the abstraction in awxkit pseudodicts.

ogajduse: >>> type(wfjt.url)
<class 'awxkit.api.pages.page.TentativePage'>

ogajduse:
Works for me :+1:
```
[INFO 200915 16:05:06] Waiting for job: 
    API: https://my.at.com/api/v2/workflow_jobs/1507/
    UI: https://my.at.com/#/workflows/1507
```

broker/providers/ansible_tower.py
mshriver: Exactly!

JacobCallahan (approved):
ACK. +1 for the UI link

Issue #44, opened 2020-08-26
Identify why test_mp_checkout is failing and how to fix it


JacobCallahan:
tests/test_broker.py test_mp_checkout is currently failing since the multiprocessing queue hangs on get.
This works perfectly fine when using the AnsibleTower provider, but this test uses the TestProvider.

Alternative solutions tried, but failed are:
- Use starmap to get around a queue
- Introduce a time delay in TestProvider's test_action method
- Use a multiprocessing manager queue
- Use a multiprocessing manager list
- Make the Host class instances pickleable to be compatible with above

jaryn:
The process child processes are zombies:
```
jhenner  3347498  3.2  0.5 296956 42672 pts/1    S    01:50   0:00 /home/jhenner/work/sat/broker/venv/bin/python venv/bin/py.test tests -v
jhenner  3347507  0.0  0.0      0     0 pts/1    Z    01:50   0:00 [py.test] <defunct>
jhenner  3347508  0.0  0.0      0     0 pts/1    Z    01:50   0:00 [py.test] <defunct>
```

The parent hangs when dealing with the Queue
```
(gdb) py-bt
Traceback (most recent call first):
  <built-in method read of module object at remote 0x7fd1114b9680>
  File "/usr/lib64/python3.9/multiprocessing/connection.py", line 384, in _recv
    chunk = read(handle, remaining)
  File "/usr/lib64/python3.9/multiprocessing/connection.py", line 419, in _recv_bytes
    buf = self._recv(4)
  File "/usr/lib64/python3.9/multiprocessing/connection.py", line 221, in recv_bytes
    buf = self._recv_bytes(maxlength)
  File "/usr/lib64/python3.9/multiprocessing/queues.py", line 103, in get
    res = self._recv_bytes()
  File "/home/jhenner/work/sat/broker/broker/broker.py", line 75, in _mp_split
    result.append(mp_queue.get())
  File "/home/jhenner/work/sat/broker/broker/broker.py", line 27, in wrapper
    return vmb_inst._mp_split(vmb_inst.checkout, count, **vmb_inst._kwargs)
  File "/home/jhenner/work/sat/broker/tests/test_broker.py", line 50, in test_mp_checkout
```

I started an attempt  to fix this using the concurrent.futures which I find easy to use, foolproof and enough for many situations. Right now I am solving it's complain that 
```
 E           RuntimeError: Queue objects should only be shared between processes through inheritance
```

PR #53, opened 2020-09-22
Use concurrent.futures to fix multiprocessing issues

jaryn:
This makes the behavior of broker in case of exception rised in the child much better (without this PR, there is a deadlock and possible no output on the terminal about the exception. With this PR the exception is propagated to the parent process.

There were issues that I think were arising from using a decorator that replaced the decorated method with some nested method (the `wraps` function is commonly used in that case). It seems to me such method cannot be pickled but that seemed necessary for sending it to the `concurrent.futures`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ python:
 install: pip install .[test]
 script:
     - cp broker_settings.yaml.example broker_settings.yaml
-    - pytest -v
+    - pytest -v tests
 deploy:
     provider: pypi
     user: jacobcallahan
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: true
 language: python
 python:
     - "3.8"
-install: pip install .
+install: pip install .[test]
 script:
     - cp broker_settings.yaml.example broker_settings.yaml
     - pytest -v
--- a/broker/broker.py
+++ b/broker/broker.py
@@ -1,10 +1,10 @@
-import multiprocessing
-from functools import wraps
 from logzero import logger
 from broker.providers.ansible_tower import AnsibleTower
 from broker.providers.test_provider import TestProvider
 from broker.hosts import Host
 from broker import helpers
+from concurrent.futures import ProcessPoolExecutor, as_completed
+from typing import Type
 
 
 PROVIDERS = {"AnsibleTower": AnsibleTower, "TestProvider": TestProvider}
@@ -16,38 +16,50 @@ PROVIDER_ACTIONS = {
 }
 
 
-def mp_split(func, count, *args, **kwargs):
-    """Split a action into multiple processes"""
-    mp_queue = multiprocessing.Queue()
-    mp_proc, result = [], []
+class mp_decorator:
+    """This decorator wraps VMBroker methods to enable multiprocessing
 
-    def concurrency_shim():
-        res = func(*args, **kwargs)
-        mp_queue.put(res)
+    The decorated method is expected to return an itearable.
+    """
 
-    for _ in range(count):
-        p = multiprocessing.Process(target=concurrency_shim)
-        mp_proc.append(p)
-        p.start()
-    for proc in mp_proc:
-        result.extend(mp_queue.get())
-    for proc in mp_proc:
-        proc.join()
-    return result
+    # Note that this is a descriptor as the other option -- using nested function
+    # like this:
+    #
+    # def mp_decorator(func)
+    #   @wraps(func)
+    #   def wrapper(func)
+    #       return
+    #
+    #   return wrapper
+    #
+    # is not working with pickling that is necessary for the ProcessPoolExecutor of
+    # concurrent.futures. I got errors like:
+    # _pickle.PicklingError: Can't pickle ... it's not the same object as ...
 
+    MAX_WORKERS = 5
+    EXECUTOR = ProcessPoolExecutor
 
-def mp_decorator(func):
-    """This decorator wraps VMBroker methods to enable multiprocessing"""
+    def __init__(self, func: callable = None):
+        self.func = func
 
-    @wraps(func)
-    def wrapper(vmb_inst: 'VMBroker', *args, **kwargs):
-        if "_count" in vmb_inst._kwargs:
-            count = vmb_inst._kwargs["_count"]
-            return mp_split(func, count, vmb_inst, *args, **kwargs)
-        else:
-            return func(vmb_inst, *args, **kwargs)
+    def __get__(self, instance: 'VMBroker', owner: Type['VMBroker']):
+        if not instance:
+            return self.func
 
-    return wrapper
+        def mp_split(*args, **kwargs):
+            count = instance._kwargs.get("_count", None)
+            if count is None:
+                return self.func(instance, *args, **kwargs)
+
+            results = []
+            with self.EXECUTOR(max_workers=self.MAX_WORKERS) as workers:
+                completed_futures = as_completed(workers.submit(self.func, instance, *args, **kwargs)
+                                                 for _ in range(count))
+                for f in completed_futures:
+                    results.extend(f.result())
+            return results
+
+        return mp_split
 
 
 class VMBroker:
@@ -101,14 +113,14 @@ class VMBroker:
                 helpers.update_inventory(add=host.to_dict())
         return hosts
 
-    def checkout(self):
+    def checkout(self, connect=False):
         """checkout one or more VMs
 
         :param connect: Boolean whether to establish host ssh connection
 
         :return: Host obj or list of Host objects
         """
-        hosts = self._checkout()
+        hosts = self._checkout(connect=connect)
         self._hosts.extend(hosts)
         return hosts if not len(hosts) == 1 else hosts[0]
 
--- a/broker/broker.py
+++ b/broker/broker.py
@@ -37,7 +37,7 @@ class mp_decorator:
 
     MAX_WORKERS = None
     """ If set to integer, the count of workers will be limited to that amount.
-     If set to None, the max workers count of the EXECUTOR will be matching the count of items."""
+     If set to None, the max workers count of the EXECUTOR will match the count of items."""
 
     EXECUTOR = ProcessPoolExecutor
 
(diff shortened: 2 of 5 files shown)


broker/providers/test_provider.py
```diff
@@ -37,7 +37,10 @@ def _set_attributes(self, host_inst, broker_args=None):
     def construct_host(self, provider_params, host_classes, **kwargs):
         host_params = provider_params.copy()
         host_params.update(kwargs)
-        host_inst = host_classes[host_params["host_type"]](**host_params)
+        host_type = host_classes[host_params["host_type"]]
+        host_type.connect = MagicMock()
```
JacobCallahan, line 41: have you considered using the MockStub class from helpers here?

jaryn: It is actually not needed to mock something here. The connect got called due to a bug in my PR which I have fixed now.

I like to use MagicMock though. It is in standard library, quite well documented and thus it is likely the other developers will know how to use it. I did use it to make the code raise an exception in `broker._act` in other tests.

.travis.yml
```diff
@@ -2,10 +2,10 @@ sudo: true
 language: python
 python:
     - "3.8"
-install: pip install .
+install: pip install .[test]
```
jaryn, line 5: I am not sure this is the best way to install the test deps. Perhaps we should just use tox.

broker/broker.py
```diff
@@ -16,21 +16,47 @@
...
+    #       return
+    #
+    #   return wrapper
+    #
+    # is not working with pickling that is necessary for the ProcessPoolExecutor of
+    # concurrent.futures. I got errors like:
+    # _pickle.PicklingError: Can't pickle ... it's not the same object as ...
 
-    return wrapper
+    MAX_WORKERS = 5
```
jaryn, line 36: I need to discuss this. I think the count of workers should be independent from the number of the machines to check out.

I also think the number of machines to check out should not be an attribute of the broker.

JacobCallahan: What benefit would there be to decoupling the number of workers from the number of vms?

It is an attribute of broker, as (almost) all broker arguments are. This is a design choice to make broker as flexible as possible.

.travis.yml
JacobCallahan: what test deps?

broker/broker.py
```diff
@@ -16,21 +16,47 @@
...
+    EXECUTOR = ProcessPoolExecutor
+
+    def __init__(self, func: callable = None):
```
JacobCallahan, line 39: no type annotations

JacobCallahan (changes requested):

.travis.yml
jaryn: https://github.com/SatelliteQE/broker/pull/53/files#diff-2eeaed663bd0d25b7e608891384b7298R12

broker/broker.py
jaryn: If we are talking about dealing with counts of VMs in ballpark figures like 20 at most then I guess we don't need to limit the workers count (it can be set to the count of the objects). We can add the limit in future if there was a need for this.

I don't understand the design choice and how it is making the broker more flexible. I don't mean this is not right, but I just don't understand.

JacobCallahan: we can have a larger discussion about how broker is designed via gchat

broker/broker.py
```diff
@@ -62,42 +87,38 @@ def _act(self, provider, method, checkout=False):
...
+        hosts = []
         for action in self._provider_actions.keys():
             provider, method = PROVIDER_ACTIONS[action]
             logger.info(f"Using provider {provider.__name__} to checkout")
             host = self._act(provider, method, checkout=True)
+            logger.debug(f"{host=} {connect=}")
             if host:
                 if connect:
                     host.connect()
-                self._hosts.append(host)
```
JacobCallahan, line 97: why are you removing this?

broker/broker.py
```diff
@@ -62,42 +87,38 @@ def _act(self, provider, method, checkout=False):
...
-            result.append(mp_queue.get())
-        for proc in mp_proc:
-            proc.join()
-        if "checkout" in method.__name__:
-            self._hosts.extend(result)
-        return result
-
     @mp_decorator
-    def checkout(self, connect=False, **kwargs):
+    def _checkout(self, connect=False):
```
JacobCallahan, line 91: what is the purpose of making this a private method, then moving existing functionality from this to the new public method?

broker/providers/test_provider.py
```diff
@@ -38,7 +36,8 @@ def _set_attributes(self, host_inst, broker_args=None):
     def construct_host(self, provider_params, host_classes, **kwargs):
         host_params = provider_params.copy()
         host_params.update(kwargs)
-        host_inst = host_classes[host_params["host_type"]](**host_params)
+        host_type = host_classes[host_params["host_type"]]
+        host_inst = host_type(**host_params)
```
JacobCallahan, line 40: why separate these two lines?

tests/test_broker.py
```diff
@@ -42,9 +43,39 @@ def test_broker_e2e():
...
+
+    cycle()
+    cycle()
+
+
+class SomeException(Exception):
+    pass
+
+
+def test_mp_checkout_exc(mocker):
```
JacobCallahan, line 74: what is the value of this test? _act is already hit multiple times in previous tests.

JacobCallahan (changes requested):

broker/broker.py
jaryn: the problem is that I found cumbersome the fact that it sometimes returned a list and sometimes a host. This is inconvenient for the caller (the generalized mp_decorator).

broker/broker.py
jaryn: Because the `self._hosts` either has to be
 * a state shared between the processes -- and we need a Manager https://docs.python.org/3/library/multiprocessing.html#sharing-state-between-processes chech the Server processes there)
 * or has to be accessed only in the parent. I think our case is perfect example when we don't need any shared state like this and we can just collect the results in the parent process.

The self._hosts is filled on https://github.com/SatelliteQE/broker/pull/53/files#diff-03b5bdba9f4a407652b7f9778cddb939R120 which is the result of the mp_split decorator 
https://github.com/SatelliteQE/broker/pull/53/files#diff-03b5bdba9f4a407652b7f9778cddb939R45

broker/providers/test_provider.py
jaryn: This is just a leftover. I can make it look as before if it matters.

tests/test_broker.py
jaryn: Well it is a test for situations like in https://github.com/SatelliteQE/broker/issues/44. There I only made fixes in the product so the `test_mp_checkout` passes in normal situations.

The `test_mp_checkout_exc` is though is a negative test (I hate that term) -- designed to raise exception (`act.side_effect = SoemeException()`) in the the child and checking the system behaves well on failure.
 * it checks whether the exception is propagated to the parent process
 * If there is a deadlock, well we should see it.

https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.side_effect

jaryn: done

broker/broker.py
```diff
@@ -16,21 +15,54 @@
...
+        if not instance:
+            return self.func
+
+        def mp_split(*args, **kwargs):
+            count = instance._kwargs.get("_count", None)
+            if count is None:
+                return self.func(instance, *args, **kwargs)
+
+            results = []
+            max_workers_count = count if self.MAX_WORKERS is None else self.MAX_WORKERS
```
JacobCallahan, line 57:
```suggestion
            max_workers_count = self.MAX_WORKERS or count
```

broker/broker.py
```diff
@@ -16,21 +15,54 @@
...
+    #
+    #   return wrapper
+    #
+    # is not working with pickling that is necessary for the ProcessPoolExecutor of
+    # concurrent.futures. I got errors like:
+    # _pickle.PicklingError: Can't pickle ... it's not the same object as ...
+
+    MAX_WORKERS = None
+    """ If set to integer, the count of workers will be limited to that amount.
+     If set to None, the max workers count of the EXECUTOR will be matching the count of items."""
```
JacobCallahan, line 40:
```suggestion
     If set to None, the max workers count of the EXECUTOR will match the count of items."""
```

JacobCallahan (changes requested):
some small changes

JacobCallahan (approved):
ACK

JacobCallahan:
@mshriver do you want to weigh in as well?

.travis.yml
mshriver: I would agree with this approach, doing the same thing in a wt.core PR at the moment.

broker/broker.py
```diff
@@ -16,21 +15,54 @@
...
+            count = instance._kwargs.get("_count", None)
+            if count is None:
+                return self.func(instance, *args, **kwargs)
+
+            results = []
+            max_workers_count = self.MAX_WORKERS or count
+            with self.EXECUTOR(max_workers=max_workers_count) as workers:
+                completed_futures = as_completed(workers.submit(self.func, instance, *args, **kwargs)
+                                                 for _ in range(count))
+                for f in completed_futures:
```
mshriver, line 61: Might be code golfing here, but couldn't you directly return here instead?

```
return [f.result() for f in completed_futures]
```

broker/broker.py
```diff
@@ -62,42 +94,38 @@ def _act(self, provider, method, checkout=False):
...
 
-        :return: Host obj or list of Host objects
+        :return: List of Host objects
         """
+        hosts = []
         for action in self._provider_actions.keys():
             provider, method = PROVIDER_ACTIONS[action]
             logger.info(f"Using provider {provider.__name__} to checkout")
             host = self._act(provider, method, checkout=True)
+            logger.debug(f"{host=} {connect=}")
```
mshriver, line 110: I'm hitting an f-string exception on this line when I try to run your branch locally.  I'm not aware of any f-string usage for `<var>=` syntax, what was your intent here?

mshriver: Discussed - its an f-string feature introduced in 3.8, but we have a need to support 3.6+ so it needs to be updated here.

jaryn:
There are some problems with the inventory.yaml unlink when doing multiple checkouts. I recommend not merging before I it is known why this happens.

jaryn:
I think the problem with race condition between the children of single process is fixed by this PR. We are ready to merge @mshriver @JacobCallahan

PR #58, opened 2020-09-24
Added --version to main broker command

### LINKED ISSUE #22: [RFE] Implement `broker --version`
# Issue 

When using broker, sometimes it can be hard to figure out what `settings.yaml` file it is using. It is similar to `ansible` where `ansible.cfg` file can be specified using an EnvVar, or can be located in `pwd` or user's home dir or under `/etc/ansible`. Since there are multiple ways to specify the file they make it obvious with `ansible --version` or by specifying the file location information on stdout during every playbook execution(if ran with `-v` for verbose). It also helps user know which python executable Ansible will be using as there could be multiple python versions installed in various locations on a given system.

# Requested solution

Please implement a way for broker to provide `broker -v/--version` to help users know what `settings.yaml`, python executable and potentially what inventory file is being used by the broker. As well as `broker` executable path itself. 

This information can be useful in debugging issues when users may run into problems with broker failing to read correct settings or not finding inventory file etc. 

# Example of `ansible --version`

```
 ansible --version
ansible 2.9.7
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/kkulkarni/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.8/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.8.3 (default, May 15 2020, 00:00:00) [GCC 10.1.1 20200507 (Red Hat 10.1.1-1)]
```

# Alternative solution

Currently readme does not specify the broker needs to execute from the same directory where the settings/inventory files are. We should update readme to reflect that OR allow user to explicitly set the path for these manually, explicitly so that user could invoke broker from any location on his system as long as it is configured to look at correct settings file. Sort of like `git config --global` where we can set global config for git to use no matter where we are calling it from.

JacobCallahan:
This provides some useful information about the files broker cares about

fixes #22

my output
```
# broker --version
Version: 0.1.3
Broker Directory: /home/jake/Programming/broker
Settings File: /home/jake/Programming/broker/broker_settings.yaml
Inventory File: /home/jake/Programming/broker/inventory.yaml
Log File: /home/jake/Programming/broker/logs/broker.log
```

### DIFF
--- a/broker/commands.py
+++ b/broker/commands.py
@@ -1,3 +1,4 @@
+import broker
 import os
 import sys
 import click
@@ -5,7 +6,7 @@ import logging
 from logzero import logger
 from broker.broker import PROVIDERS, PROVIDER_ACTIONS, VMBroker
 from broker import logger as b_log
-from broker import helpers
+from broker import helpers, settings
 
 
 def update_log_level(ctx, param, value):
@@ -63,7 +64,7 @@ def populate_providers(click_group):
             )(provider_cmd)
 
 
-@click.group()
+@click.group(invoke_without_command=True)
 @click.option(
     "--log-level",
     type=click.Choice(["info", "warning", "error", "critical", "debug", "silent"]),
@@ -72,8 +73,22 @@ def populate_providers(click_group):
     is_eager=True,
     expose_value=False,
 )
-def cli():
-    pass
+@click.option(
+    "--version",
+    is_flag=True,
+    help="Get broker system-level information",
+)
+def cli(version):
+    if version:
+        with open("setup.py") as setup_file:
+            for line in setup_file:
+                if "version" in line:
+                    click.echo(f"""Version: {line.split('"')[1]}""")
+        broker_directory = settings.BROKER_DIRECTORY.absolute()
+        click.echo(f"Broker Directory: {broker_directory}")
+        click.echo(f"Settings File: {settings.settings_path.absolute()}")
+        click.echo(f"Inventory File: {broker_directory}/inventory.yaml")
+        click.echo(f"Log File: {broker_directory}/logs/broker.log")
 
 
 @cli.command(


kedark3 (approved):

broker/commands.py
```diff
@@ -72,8 +73,22 @@ def provider_cmd(*args, **kwargs):  # the actual subcommand
...
+    "--version",
+    is_flag=True,
+    help="Get broker system-level information",
+)
+def cli(version):
+    if version:
+        with open("setup.py") as setup_file:
+            for line in setup_file:
+                if "version" in line:
+                    click.echo(f"""Version: {line.split('"')[1]}""")
```
mshriver, line 86: Skeptical about this being stable without being a more specific match (regex), as even any comments in setup.py about the version might break it.

I'm at the moment not sure of a better way to do this that would work for running broker installed from pip, and installed from a local clone. If we started tagging, we could use `git describe --tag`, but that only works for local install.  Reading from pip would work too, but isn't going to give you a meanginful version with a local install.

mshriver (approved):

broker/commands.py
JacobCallahan: unless setup.py standards change, this should remain stable. Consider the line itself
`    version="0.1.3",`
splitting that by `"` results in a 3 part list
`['    version=', '0.1.3', ',']`
this will always be the case since the version argument is a string. the only way this could fail in a foreseeable future would be if someone changed this line from double to single quote strings.

mshriver: I'm more so talking about a situation like:

```
setup@localhost:~/repos/broker (1029382$*) % git diff setup.py
diff --git a/setup.py b/setup.py
index 92c5958..1f07c58 100644
--- a/setup.py
+++ b/setup.py
@@ -11,6 +11,7 @@ requirements = ["awxkit", "click", "dynaconf>=3.1.0", "logzero", "pyyaml", "ssh2
 
 setup(
     name="broker",
+    # the version is "variable"
     version="0.1.3",
     description="The infrastructure middleman.",
     long_description=readme + "\n\n" + history,

setup@localhost:~/repos/broker (1029382$*) % broker --version
Version: variable
Version: 0.1.3
```

It's not major, just pointing out its a loose match. Might end up throwing some extra info if we aren't careful about what's put in setup.py.

JacobCallahan: haha, I get you. I'll assume that risk for now :)

PR #59, opened 2020-09-24
Don't run pytest on other dir than tests

jaryn:
This is to work arround issue with TestProvider class attempts to be collected
 as test class.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,13 @@
 sudo: true
 language: python
 python:
+    - "3.6"
+    - "3.7"
     - "3.8"
 install: pip install .[test]
 script:
     - cp broker_settings.yaml.example broker_settings.yaml
-    - pytest -v
+    - pytest
 deploy:
     provider: pypi
     user: jacobcallahan
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ python:
 install: pip install .[test]
 script:
     - cp broker_settings.yaml.example broker_settings.yaml
-    - pytest -v tests
+    - pytest -v
 deploy:
     provider: pypi
     user: jacobcallahan
--- a/broker/providers/test_provider.py
+++ b/broker/providers/test_provider.py
@@ -14,6 +14,8 @@ HOST_PROPERTIES = {
 
 
 class TestProvider(Provider):
+    __test__ = False
+
     def __init__(self, **kwargs):
         self.config = settings.TESTPROVIDER.config_value
 
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,3 @@
+[pytest]
+testpaths = tests
+addopts = -v
--- a/setup.py
+++ b/setup.py
@@ -39,6 +39,8 @@ setup(
         "License :: OSI Approved :: GNU General Public License v3 (GPLv3)",
         "Natural Language :: English",
         "Programming Language :: Python :: 3",
+        "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
     ],
 )


RonnyPfannschmidt:
i believe it would be a help to instead set https://docs.pytest.org/en/3.4.2/customize.html#confval-testpaths as it would work for local as well

RonnyPfannschmidt:
additionally it may be helpful to mark TestProvider as not a test in general

jaryn:
> additionally it may be helpful to mark TestProvider as not a test in general

I did want to do that, but haven't found a way other than renaming the `TestProvider.` But I didn't look too hard.

RonnyPfannschmidt:
Pytest adheres to `__test__ = False`, i believe it's usable in that case

mshriver:
@jaryn I like @RonnyPfannschmidt suggestion to use `testpaths` in pytest.ini to achieve the original comment of limiting test collection. That way collection is limited for local testing and CI.   Can you include that in this PR as well?

jaryn:
> Pytest adheres to `__test__ = False`, i believe it's usable in that case

@JacobCallahan @mshriver  Why is the `TestProvider` outside of the `tests` directory anyway? It seems it is not used in the CLI because of this filter: https://github.com/SatelliteQE/broker/blob/master/broker/commands.py#L43 so I think the only use is for the tests. I wonder why is this  system not done in more pluggable fashion.

jaryn:
> @jaryn I like @RonnyPfannschmidt suggestion to use `testpaths` in pytest.ini to achieve the original comment of limiting test collection. That way collection is limited for local testing and CI. Can you include that in this PR as well?

 * Well I don't know, but when all the modules are tried by py.test, doesn't it mean their importability is actually tested?
 * Also what should I include? If we were to allow doctest, than we would need the broker dir being collected and that means we are collecting most of the present content of this repo anyway.

```
[pytest]
testpaths = tests broker
```

JacobCallahan:
@jaryn TestProvider lives in the providers directory because it is a provider. It can be used from the CLI, it just is filtered from the providers listed in that command to not confuse CLI users.
How does TestProvider being in the providers directory make broker less pluggable?

.travis.yml
```diff
@@ -1,11 +1,13 @@
 sudo: true
 language: python
 python:
+    - "3.6"
```
jaryn, line 4: In https://github.com/SatelliteQE/broker/pull/53#discussion_r499669068 is stated we need to support Python 3.6

jaryn:
> @jaryn TestProvider lives in the providers directory because it is a provider. It can be used from the CLI, it just is filtered from the providers listed in that command to not confuse CLI users.
> How does TestProvider being in the providers directory make broker less pluggable?

jaryn:
> @jaryn TestProvider lives in the providers directory because it is a provider. It can be used from the CLI, it just is filtered from the providers listed in that command to not confuse CLI users.
> How does TestProvider being in the providers directory make broker less pluggable?

Well it seems to me it is not just filtered from listing the providers:
```
$ broker providers 
Usage: broker providers [OPTIONS] COMMAND [ARGS]...

  Get information about a provider and its actions

Options:
  --help  Show this message and exit.

Commands:
  AnsibleTower  Get information about a provider's actions
$ broker providers AnsibleTower
[INFO 201013 13:41:13] Querying provider AnsibleTower
[INFO 201013 13:41:13] Using username and password authentication
```

So AnsibleTower can be used, but:
```
$ broker providers TestProvider
Usage: broker providers [OPTIONS] COMMAND [ARGS]...
Try 'broker providers --help' for help.

Error: No such command 'TestProvider'.
```

Evidently cannot. Therefore I conclude the TestProvider cannot be used from CLI without manually changing the code in *commands.py*.

I think the TestProvider is only used as mock for tests and thus should reside in the tests dir.
If it can be used from CLI and should be hidden, I think this is the way to do it:
https://click.palletsprojects.com/en/7.x/api/?highlight=hidden#click.Option here is an example:
https://github.com/SatelliteQE/broker/pull/69

mshriver (approved):

PR #61, opened 2020-09-29
Better discovery

jaryn:


mshriver:
I like the approach here, though I'm not sure that I've fully wrapped my head around how the decorator is behaving during `Provider._act`. I'll have to step through things to be more confident in approving the change.

I support the idea of using entrypoints to register providers, and using a decorator to identify provider actions, instead of data structures at the top of the module.

PR #62, opened 2020-09-29
Clenups

jarovo:


broker/broker.py
```diff
@@ -104,10 +108,7 @@ def execute(self, **kwargs):
 
         :return: Any The results given back by the provider
```
swadeley, line 109:
```suggestion
        :return: Any results given back by the provider
```

broker/providers/__init__.py
```diff
@@ -8,21 +9,31 @@ def __init__(self):
...
-        return {k: v for k, v in kwargs.items() if k in arg_list}
+    def _set_attributes(self, host_inst, broker_args=None):
+        host_inst.__dict__.update(
+            {
+                "release": self._host_release,
+                "_test_inst": self,
+                "_broker_provider": "TestProvider",
+                "_broker_args": broker_args,
+            }
+        )
```
JacobCallahan, line 20: This shouldn't have any reference to a child provider, in this case TestProvider. 

To make this flexible, you could switch it so `_broker_provider` is set to `self.__class__.__name__`

broker/providers/__init__.py
```diff
@@ -1,4 +1,5 @@
 import inspect
+from abc import ABC, abstractmethod
```
JacobCallahan, line 2: were you planning on using these?

broker/providers/ansible_tower.py
```diff
@@ -26,7 +26,7 @@
 
 class AnsibleTower(Provider):
     def __init__(self, **kwargs):
-        self._construct_params = []
+        super().__init__()
```
JacobCallahan, line 29: ehhhh... I'm not sure if the calls are worth it at this point.

broker/providers/ansible_tower.py
```diff
@@ -64,16 +64,6 @@ def _host_release(self):
...
-    def _set_attributes(self, host_inst, broker_args=None):
-        host_inst.__dict__.update(
-            {
-                "release": self._host_release,
-                "_at_inst": self,
-                "_broker_provider": "AnsibleTower",
-                "_broker_args": broker_args,
-            }
-        )
-
```
JacobCallahan, line 76: In this PR's current state, the _broker_provider would become TestProvider

broker/providers/test_provider.py
```diff
@@ -25,22 +26,11 @@ def _host_release(self):
...
+    def nick_help(self, **kwargs):
+        pass
 
     def construct_host(self, provider_params, host_classes, **kwargs):
-        host_params = provider_params.copy()
-        host_params.update(kwargs)
-        host_inst = host_classes[host_params["host_type"]](**host_params)
-        self._set_attributes(host_inst, broker_args=kwargs)
-        return host_inst
+        return super().construct_host(provider_params, host_classes, **kwargs)
```
JacobCallahan, line 33: with this PR, you wouldn't need to define `construct_host` here

JacobCallahan (changes requested):

PR #64, opened 2020-10-05
Remove PyPi secrets from git, Remove Travis, use Github Actions and Py3.6 Py3.7 testing

jaryn:
* This removes support of travis and adds use of Github Actions
 * I am also introducing a support for github actions as an experiment.

Should we build on the pbr package? It manages some things, for example the requriements.txt:
https://docs.openstack.org/pbr/3.0.0/#requirements
But on the other hand, https://packaging.python.org/discussions/install-requires-vs-requirements/ says something in the regards of the `install_requires` in `setup.py` should be specifying only some ranges that should be working with (thus allowing better interoperability with other packages), while `requirements.txt` is for pinning the version for reproducibility.


mshriver:
Not to burst any bubbles, but I'd like to keep broker off of tox.

I think we can target python 3.8 for CI, adding new versions as we want to test them in CI.

I like your approach of adding an extras_require for pip install with `[test]` to get the dependencies for testing locally.

Are you open to changing your focus/approach here to moving us to github workflows for running pre-commit/pytest, instead of using travis?

jaryn:
> Not to burst any bubbles, but I'd like to keep broker off of tox.
Not sure why, but I am not insisting on using tox. I just found it useful to be able to run the tests locally.
> 
> I think we can target python 3.8 for CI, adding new versions as we want to test them in CI.
> 
> I like your approach of adding an extras_require for pip install with `[test]` to get the dependencies for testing locally.
OK. I think it is included in this PR. But probably should.
> 
> Are you open to changing your focus/approach here to moving us to github workflows for running pre-commit/pytest, instead of using travis?

I did add github workflows as well as I've changed the travis.yaml. They are now both passing (The github actions on my personal repo and the Sattelite travis).

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,41 @@
...
+name: Python package
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
```
mshriver, line 13: Can you give this a bit more descriptive/pretty name, since its not really building, but running pytest?

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,41 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        cp broker_settings.yaml.example broker_settings.yaml
+        python -m pip install --upgrade pip
```
mshriver, line 29: With the setup-python action you shouldn't really have to worry about upgrading pip.

I think it would be best to have a 'test' extra_requires for setuptools, and then just `pip install -e .[test]` here, instead of looking for requirements files.

tox.ini
```diff
@@ -0,0 +1,8 @@
+[tox]
```
mshriver, line 1: Please remove tox and just use pytest directly within the github workflow. Matrix for the workflow should take care of running against multiple versions of python.

.travis.yml
```diff
@@ -1,11 +1,13 @@
 sudo: true
 language: python
 python:
+    - "3.6"
```
mshriver, line 4: We're not concerned about testing versions earlier than 3.8, but since 3.9 just released it would be great to include that.

.github/workflows/python-package.yml
jaryn: I will try. I have used pre-existing scaffold that seemed to do what we want.

.github/workflows/python-package.yml
jaryn: Upgrading pip was part of the scaffold. Not sure is it safer to upgrade or not-upgrade it every time, but I am certain it is not necessary. I will remove that.

> I think it would be best to have a 'test' extra_requires for setuptools, and then just pip install -e .[test] here, instead of looking for requirements files.

I think you mean to have setuptools among the the extra_requires. OK, will do.

How are the setuptools related to requirements file I don't know, but the file is not present and IIUC we don't need it unless we are composing an reproducible environment of mulitiple packaged projects.

.travis.yml
jaryn: https://github.com/SatelliteQE/broker/pull/53#discussion_r499669068 tells we need support 3.6+.  So according to that I am not sure if 3.6, but I am certain it means 3.7 is meant to be supported and therefore the removing the f-string debug was justified.

Are we or are we not concerned about <3.8 and thus can I use f-string debugs?

tox.ini
jaryn: not when running locally, but ok

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,31 @@
...
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install setuptools wheel twine
+    - name: Build and publish
```
mshriver, line 25: I think we should use pypa's action for this:
https://github.com/pypa/gh-action-pypi-publish

mshriver (commented):
I'd support this move to github actions.

Concerning use of install_requires / pbr+requirements.txt, I would leave it as is with install_requires and loose versions where possible.

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,39 @@
...
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  tests:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.7, 3.8]
```
JacobCallahan, line 18: 3.8 and 3.9

JacobCallahan (changes requested):

mshriver:
I've taken these commits and made some further updates, not including tox anymore:
https://github.com/SatelliteQE/broker/pull/116

PR #68, opened 2020-10-12
Minor fix for AnsibleTower provider

JacobCallahan:
This accounts for when tower doesn't provide a child node result


mshriver (approved):
Ever have deja vu?

PR #71, opened 2020-10-23
Adding a broker checkout and checkin sample library

kedark3:
@JacobCallahan Adding this code here as you requested. Kindly provide inputs so that this matches what you may have had in your mind on how we want to serve this content as example through broker repo.

### DIFF
--- a/jenkins_example/brokerUtils.groovy
+++ b/jenkins_example/brokerUtils.groovy
@@ -1,0 +1,49 @@
+//Helper functions to spin up new Satellite/Capsule/RHEL instances
+
+def checkout(Map parameters = [:]) {
+    /*
+    Here is an example of how one could invoke this method:
+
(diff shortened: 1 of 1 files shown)


JacobCallahan:
Not sure if this is the best place for it, but we can discuss that more next week.

mshriver:
I'm hesitant to include an example like this in broker.  Its not really related to broker, which provides both a Dockerfile for container building (and we can publish an upstream container on docker and quay) and is pip installable.  The Jenkins example provided works for us but is pretty dependent on the environment of the jenkins node executing the groovy, and doesn't really exemplify anything about broker, just some groovy basics for running shell commands with dynamic variables.

kedark3:
@mshriver  this was opened as @JacobCallahan wanted to add it here. If that's no longer the case, feel free to close :)

JacobCallahan:
@kedark3 the intent from the requester was more to have it open sourced in general. The exact place it should live, I am not sure. Maybe it should live in a collections type repo.

PR #72, opened 2020-10-30
Added a list-templates ability for AnsibleTower provider

JacobCallahan:
list-templates becomes a first-class component of AnsibleTower
new setting added to limit the amount of templates shown. default: 20

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,6 +1,14 @@
 History
 =======
 
+0.1.5 (2020-10-31)
+------------------
+
++ Added a list-templates ability for AnsibleTower provider
++ Added GNU license
++ minor fixes
+- removed Michael
+
 0.1.4 (2020-10-12)
 ------------------
 
--- a/broker/providers/ansible_tower.py
+++ b/broker/providers/ansible_tower.py
@@ -3,6 +3,7 @@ import json
 import sys
 from urllib import parse as url_parser
 from broker.settings import settings
+from broker.helpers import results_filter
 from logzero import logger
 from datetime import datetime
 
@@ -39,7 +40,9 @@ class AnsibleTower(Provider):
         if TOKEN:
             logger.info("Using token authentication")
             config.token = TOKEN
-            root.connection.login(username=None, password=None, token=TOKEN, auth_type='Bearer')
+            root.connection.login(
+                username=None, password=None, token=TOKEN, auth_type="Bearer"
+            )
             versions = root.get().available_versions
             try:
                 # lookup the user that authenticated with the token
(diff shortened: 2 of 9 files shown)


HISTORY.md
```diff
@@ -1,6 +1,14 @@
 History
 =======
 
+0.1.5 (2020-10-30)
+------------------
+
++ Added a list-templates ability for AnsibleTower provider
++ Added GNU license
++ minor fixes
+- removed Michael
```
mshriver, line 10: well that's not nice

broker/providers/ansible_tower.py
```diff
@@ -250,16 +276,27 @@ def extend_vm(self, target_vm):
...
             workflows = self.v2.workflow_job_templates.get().results
             workflows = "\n".join([workflow.name for workflow in workflows])
             logger.info(f"Available workflows:\n{workflows}")
+        elif kwargs.get("templates"):
+            templates = self.exec_workflow(workflow="list-templates", artifacts="last")
+            logger.info(settings.ANSIBLETOWER.template_limit)
+            templates = "\n".join(
+                tmpl
+                for tmpl in templates["data_out"]["list_templates"][
+                    : settings.ANSIBLETOWER.template_limit
```
mshriver, line 294: I think it would be good to be able to pass this, not to have to modify envvar or broker_settings.yaml when you want to adjust how many templates are returned. Without filtering/sorting of the templates here (yet) it could take a few re-runs until the template someone is looking for is shown.

mshriver (approved):
Great support and start, there's a few things I think we'll need to add to this behavior to add functionality, like sorting/filtering the template list.

HISTORY.md
JacobCallahan: wrong michael :D

broker/providers/ansible_tower.py
JacobCallahan: added

JacobCallahan:
In the latest changes, I also added the ability to limit and filter results from the CLI

latran:
1. I'm assuming the list-templates lists whatever satlab has regardless of what sat version it is, correct?
2. Do you think it's worthwhile to have a `list-template count` that shows a count for all templates available for a sat version or all sat versions (6.8, 6.7, etc)?
3. Can you provide some usage sample commands for the filter portion?

JacobCallahan:
@latran thanks for the questions!
1. Correct. satlab specifically will list all available templates, regardless of the contents.
2. If you think this is valuable, you could submit a PR for it. On the other hand, that can be gained via normal linux grep/head/etc...
3. filters are detailed in-depth in Broker's readme. The only difference is that the "key" of the filter doesn't matter for a results-filter, meaning you can put any name there.

PR #73, opened 2020-11-03
Fix broker version command

JacobCallahan:
It currently reads from a relative setup.py file
New change doesn't rely on setup.py
Also small changes to providers generated help, removing duplicated limit and filter entires
Removed duplicates from --list-templates and reverse sorted ("higher" versions first)

### DIFF
--- a/broker/commands.py
+++ b/broker/commands.py
@@ -62,12 +62,12 @@ def populate_providers(click_group):
             provider_cmd = click.option(
                 f"--{action}", type=str, help=f"Get information about a {action}"
             )(provider_cmd)
-            provider_cmd = click.option(
-                f"--results-limit", type=int, help=f"The maximum number of results to get back"
-            )(provider_cmd)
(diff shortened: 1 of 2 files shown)


broker/commands.py
```diff
@@ -86,10 +86,9 @@ def provider_cmd(*args, **kwargs):  # the actual subcommand
 )
 def cli(version):
     if version:
-        with open("setup.py") as setup_file:
-            for line in setup_file:
-                if "version" in line:
-                    click.echo(f"""Version: {line.split('"')[1]}""")
+        import pkg_resources
+        broker_version = pkg_resources.get_distribution("broker").version
```
mshriver, line 90: :ok_hand:

mshriver (approved):

broker/commands.py
JacobCallahan: yeah this is a much better method!

lpramuk (approved):

PR #75, opened 2020-11-16
small PR for merge_artifact

latran:
Changing strategy="merge" to strategy="last"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 sudo: true
 language: python
 python:
-    - "3.6"
-    - "3.7"
     - "3.8"
 install: pip install .[test]
 script:
--- a/broker/providers/ansible_tower.py
+++ b/broker/providers/ansible_tower.py
@@ -92,7 +92,7 @@ class AnsibleTower(Provider):
             }
         )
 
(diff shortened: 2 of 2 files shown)


JacobCallahan (approved):
ACK. So the intent of this is to allow Tower administrators to use jobs that a user doesn't have permission to view in the workflow execution process. As long as they make the last job (that presents results) viewable by the user, broker will still process them correctly.

mshriver:
@JacobCallahan Since we're using py3.8 features, we need to turn 3.6 and 3.7 of for travis.  @latran want to include that in this PR, its just removing two lines from `.travis.yml`?

latran:
@mshriver, @JacobCallahan Ok done.

.travis.yml
```diff
@@ -1,8 +1,6 @@
 sudo: true
 language: python
 python:
-    - "3.6"
```
mshriver, line 4: Thanks for including this!

latran: @mshriver ?.  Are you saying thanks for including the removal? If so, English is weird.

mshriver: yes and yes, it sure is weird.

mshriver (approved):

PR #78, opened 2020-12-08
Populate missing field values based on returned results from Ansible …

JacobCallahan:
…Tower

This change allows broker to read the extra_vars in a workflow and
attempt to populate values in them from results returned back from
tower.
This will provide a more consistent behavior when duplicating a host.
Additionally, the new information can let users know exactly what they
got back from workflows that allow minimal or no arguments.

### DIFF
--- a/broker/providers/ansible_tower.py
+++ b/broker/providers/ansible_tower.py
@@ -162,26 +162,26 @@ class AnsibleTower(Provider):
             "type": host.type,
             "hostname": host.variables["fqdn"],
             "_broker_provider": "AnsibleTower",
+            "_broker_args": getattr(host, "_broker_args", {})
         }
         expire_time = self._get_expire_date(host.id)
         if expire_time:
             host_info["expire_time"] = expire_time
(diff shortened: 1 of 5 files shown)


broker/providers/ansible_tower.py
```diff
@@ -199,6 +199,12 @@ def construct_host(self, provider_params, host_classes, **kwargs):
             job_attrs = self._merge_artifacts(
                 job, strategy=kwargs.get("strategy", "merge")
             )
+            # pull information about the job arguments
+            job_extra_vars = json.loads(job.extra_vars)
+            # and update them if they have resolved values
+            for key in job_extra_vars.keys():
+                job_extra_vars[key] = job_attrs.get(key)
```
mshriver, line 206: Is this just equivalent to `job_extra_vars.update(job_attrs)`?

mshriver: Talked offline - nope! because of the behavior of .get setting None

mshriver (approved):

PR #79, opened 2021-01-15
Added ability to execute job templates

JacobCallahan:
Broker can now execute and query AnsibleTower job templates
job templates should also be able to be abstracted via nicks

```
broker providers AnsibleTower --job-templates
broker providers AnsibleTower --job-template <jt name>

broker execute --job-template <jt-name>
```


JacobCallahan:
@bherrin3 here it is

bherrin3:
## Additional Context

Adding some context. :)

## User Story

Moving Ansible Tower meta into external repos means we need the ability to test that meta on MR/PRs or at least some cron to verify integrity of the repo from the infecting code

## broker request for job_template calls

All job_templates should have some pre-made input values that would test to verify the functionality of the job_template. Therefore, an easy way to do this is to be able to use `broker` in CI to call the AnsibleTower API's JT with the appropriate `extra_vars` to verify a MR/PR has not broken the functionality. 

Adding this functionality enables `broker` to be used within MR/PR testing for this lower-order use case.

For Workflows, this is already covered by the current `broker` functionality

## Items outside the scope

For playbooks and roles, `molecule` will be used and is outside the scope of this project.

## Additional Follow-up 

Will be following up with Sean and tower_configuration to ensure this is not duplicating a solution that may already exist.

broker/providers/ansible_tower.py
```diff
@@ -224,39 +224,44 @@ def construct_host(self, provider_params, host_classes, **kwargs):
...
         )
         job.wait_until_completed(timeout=AT_TIMEOUT)
         if not job.status == "successful":
             logger.error(
-                f"Workflow Status: {job.status}\nExplanation: {job.job_explanation}"
+                f"{subject.capitalize()} Status: {job.status}\nExplanation: {job.job_explanation}"
             )
             return
-        if artifacts:
+        if (artifacts := kwargs.pop("artifacts")):
```
mshriver, line 264: pop will raise `AttributeError` here, as seen in unit tests.

JacobCallahan: s/pop/get will fix that!

broker/commands.py
```diff
@@ -279,6 +280,7 @@ def duplicate(vm, background, count, all_, filter):
 )
 @click.option("-b", "--background", is_flag=True, help="Run execute in the background")
 @click.option("--workflow", type=str)
+@click.option("--job-template", type=str)
```
mshriver, line 283: workflow and job-template should be mutually exclusive options. It would take a bit of rework for handling the kwargs in `execute`, but it would be good to represent this with the command options. `click.Choice` appears to be a good option -- but would force a change in the pattern of the CLI use, specifying a _type_ of execution as a key/value. 

This looks fairly promising if we want to add a dependency for a click plugin:
https://github.com/click-contrib/click-option-group

Non blocking for this implementation, but its introducing a bit of ambiguity to the CLI options.  Adding help text to both options explaining their mutual exclusivity should point users in the right direction.

broker/providers/ansible_tower.py
```diff
@@ -224,39 +224,45 @@ def construct_host(self, provider_params, host_classes, **kwargs):
...
+            return
+        candidates = get_path.get(name=name).results
+        if candidates:
+            target = candidates.pop()
+        else:
+            logger.error(f"{subject.capitalize()} not found by name: {name}")
             return
         logger.debug(
-            f"Launching workflow template: {url_parser.urljoin(AT_URL, str(wfjt.url))}"
+            f"Launching workflow template: {url_parser.urljoin(AT_URL, str(target.url))}"
```
mshriver, line 250: The log message prefix should include `subject`, `f'Launching {subject}: <...>'`

JacobCallahan: good catch

broker/commands.py
JacobCallahan: I'll think about this some more as a number of other things would need to be changed if we want to carry this pattern across the cli.

mshriver: Yeah I think the click-option-group is a good option here, so we can just simply specify options that are mutually exclusive, and keep the CLI pattern the same.

mshriver (approved):

PR #84, opened 2021-02-05
List AnsibleTower Inventories

ColeHiggins2:
**Description:**

With the incoming requirement to have broker accept and pass along inventory information, it would be nice to let users be able to figure out what queries are available.

Desired usage:
Get a list of inventories your account has access to.

**broker providers AnsibleTower --inventories**

Optional usage:
Get some relevant information about an inventory.

**broker providers AnsibleTower --inventory test-inventory**

Test Results:

Screenshot from 2021-02-05 15-04-09

### DIFF
--- a/broker/commands.py
+++ b/broker/commands.py
@@ -57,8 +57,9 @@ def populate_providers(click_group):
             if prov_info[0] == prov_class
         ):
             action = action.replace("_", "-")
+            plural = action.replace('y', 'ies') if action.endswith('y') else f"{action}s"
             provider_cmd = click.option(
-                f"--{action}s", is_flag=True, help=f"Get available {action}s"
+                f"--{plural}", is_flag=True, help=f"Get available {plural}"
             )(provider_cmd)
             provider_cmd = click.option(
                 f"--{action}", type=str, help=f"Get information about a {action}"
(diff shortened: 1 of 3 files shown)


ColeHiggins2:
Fixed Inventory output results:


Screenshot from 2021-02-05 16-00-14

JacobCallahan (approved):
ACK, nice addition!

Fixes #84

broker/commands.py
```diff
@@ -57,8 +57,9 @@ def provider_cmd(*args, **kwargs):  # the actual subcommand
             if prov_info[0] == prov_class
         ):
             action = action.replace("_", "-")
+            plural = action.replace('y', 'ies') if action.endswith('y') else f"{action}s"
```
mshriver, line 60: I'm suspect of this change, we now have to be careful about adding provider actions that end in `y`.

Consider adding `inflect` as a requirement and using it for this conversion.

mshriver (approved):

broker/commands.py
JacobCallahan: Let's cross that bridge if we come to it. I'd like to not introduce more dependencies if they aren't currently necessary.

PR #89, opened 2021-02-26
Added the ability to pass in complex data structures via files

### LINKED ISSUE #87: Broker does not provide a way to pass non-trivial variables as an argument
I would like broker to accept JSON as an argument.

Let's assume the following JSON.
```json
[{"baseurl": "http://my.mirror/rpm", "name": "base"}, {"baseurl": "http://my.mirror/rpm","name": "optional"}]
```
I was not able to pass this data structure as an argument.

Users might modify their `broker_settings.yaml` and create their own nick for this purpose. However, modifying `broker_settings.yaml` is not desired workflow within the CI.

I can imagine broker being able to either accept the following:
```
broker checkout --workflow my-workflow --myarg '[{"baseurl": "http://my.mirror/rpm", "name": "base"}, {"baseurl": "http://my.mirror/rpm","name": "optional"}]' --my_second_arg foo
```

or provide an option to pass a file with the JSON data structure as an argument like so:

`myvars.json`:
```json
{"myarg":[{"baseurl": "http://my.mirror/rpm", "name": "base"}, {"baseurl": "http://my.mirror/rpm","name": "optional"}], "my_second_arg": "foo"}
```
Command using the variable file:
```
broker checkout --workflow my-workflow --vars-file myvars.json
```

JacobCallahan:
This change allows users to use files to store arguments.
Broker will attempt to resolve these files at runtime and pass them
along to the provider.
Since the values override the runtime broker_args, the resolved values
are stored in the inventory file so modified/destroyed files won't
result in improper behavior during a duplicate action.

fixes #87

### DIFF
--- a/broker/helpers.py
+++ b/broker/helpers.py
@@ -1,10 +1,15 @@
 """Miscellaneous helpers live here"""
-from collections import namedtuple, UserDict
+import json
+import sys
+from collections import UserDict, namedtuple
 from collections.abc import MutableMapping
 from copy import deepcopy
-from broker import settings
-import yaml
+from pathlib import Path
 
+import yaml
+from logzero import logger
+
+from broker import settings
 
 FilterTest = namedtuple("FilterTest", "haystack needle test")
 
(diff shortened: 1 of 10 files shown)


ogajduse (approved):
It looks great from the user's perspective. Thanks, @JacobCallahan!

@rplevka can you have a look at this PR? Will this solution work for our use case in the Jenkins pipeline and the broker container?

rplevka:
Thanks @JacobCallahan.
We can definitely use this in our pipeline (as long as we are able to create new files inside broker container)

rplevka (approved):

broker/helpers.py
```diff
@@ -134,6 +138,48 @@ def resolve_nick(nick):
         return settings.settings.NICKS[nick].to_dict()
 
 
+def load_file(file):
+    """Verifies existence and loads data from json and yaml files"""
+    file = Path(file)
+    if not file.exists() or file.suffix not in (".json", ".yaml", ".yml"):
+        logger.warning(f"File {file.absolute()} is invalid or does not exist.")
```
mshriver, line 145: I'd consider this an error - the user provided data that was invalid, the process should end at this point.

broker/helpers.py
```diff
@@ -134,6 +138,48 @@ def resolve_nick(nick):
...
+def resolve_file_args(broker_args):
+    """Check for files being passed in as values to arguments,
+    then attempt to resolve them. If not resolved, keep arg/value pair intact.
+    """
+    final_args = {}
+    for key, val in broker_args.items():
+        if isinstance(val, Path) or (
+            isinstance(val, str) and val[-4:] in ("json", "yaml", ".yml")
+        ):
+            if (data := load_file(val)) :
```
mshriver, line 167: Right now if the file was invalid or doesn't exist, this is false, and nothing is loaded from the file. I think this needs to stop execution when this condition is hit. This could otherwise start something like a checkout that takes 20 minutes and can't be stopped - and has had the wrong arguments passed.

broker/helpers.py
JacobCallahan: So this isn't something we should error out on since there are valid cases where someone could be passing in file path locations to a provider for an action. In that case, Broker would be incorrectly assuming the argument is meant for him than the provider.

broker/helpers.py
JacobCallahan: See my above comment :)

Unless we add a distinguishing mark of when a passed in file is for Broker vs a provider, then there is no way we can tell (outside of `--args-file`) if the missing data is expected or not.

broker/helpers.py
mshriver: I don't see how a given file not being present or compatible is related to the mode of operation here. In either case, I tried to provide broker or provider arguments via a file, and that file wasn't there or is the wrong type.

mshriver (approved):

PR #93, opened 2021-03-11
Fix traceback when checking out multiple VMs with connect=True

### LINKED ISSUE #92: Traceback when checking out multiple VMs with `connect=True`
Checkout fails when `_count > 1` and `connect=True`:

```
>>> from broker.broker import VMBroker
>>> vmb = VMBroker(nick='rhel7', _count=2)
>>> vms = vmb.checkout()
>>> vms
[<broker.hosts.Host object at 0x7fd4d8605ac0>, <broker.hosts.Host object at 0x7fd4d8605fa0>]
>>> vmb.checkin()

>>> vms = vmb.checkout(connect=True)
[...]
concurrent.futures.process._RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/usr/lib64/python3.9/concurrent/futures/process.py", line 208, in _sendback_result
    result_queue.put(_ResultItem(work_id, result=result,
  File "/usr/lib64/python3.9/multiprocessing/queues.py", line 372, in put
    obj = _ForkingPickler.dumps(obj)
  File "/usr/lib64/python3.9/multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
  File "stringsource", line 2, in ssh2.session.Session.__reduce_cython__
TypeError: no default __reduce__ due to non-trivial __cinit__
"""
```

This also happens when using the broker instance in a `with` block, because `VMBroker.__enter__()` calls `checkout()` with `connect=True`.

When `_count > 1`, `VMBroker._checkout()` runs the checkout tasks in parallel, using `concurrent.futures`. That module uses pickle to serialize the host instances and pass them between the worker and parent process. When `connect=True`, then `_checkout()` also establishes an ssh connection to each host and stores a reference to the un-pickle-able ssh connection instance in `host.session`. The traceback is the result of trying to pickle that attribute.

Proposed patch is to move the calls to `host.connect()` outside of `_checkout()`, which should only perform the actions that need to be run in parallel, and into the calling method `checkout()`:

```
$ git --no-pager diff --staged
diff --git a/broker/broker.py b/broker/broker.py
index 8b09943..2da3ed2 100644
--- a/broker/broker.py
+++ b/broker/broker.py
@@ -100,11 +100,9 @@ class VMBroker:
             return result
 
     @mp_decorator
-    def _checkout(self, connect=False):
+    def _checkout(self):
         """checkout one or more VMs
 
-        :param connect: Boolean whether to establish host ssh connection
-
         :return: List of Host objects
         """
         hosts = []
@@ -112,10 +110,8 @@ class VMBroker:
             provider, method = PROVIDER_ACTIONS[action]
             logger.info(f"Using provider {provider.__name__} to checkout")
             host = self._act(provider, method, checkout=True)
-            logger.debug(f"host={host} connect={connect}")
+            logger.debug(f"host={host}")
             if host:
-                if connect:
-                    host.connect()
                 hosts.append(host)
                 logger.info(f"{host.__class__.__name__}: {host.hostname}")
         return hosts
@@ -127,7 +123,10 @@ class VMBroker:
 
         :return: Host obj or list of Host objects
         """
-        hosts = self._checkout(connect=connect)
+        hosts = self._checkout()
+        if connect:
+            for host in hosts:
+                host.connect()
         self._hosts.extend(hosts)
         helpers.update_inventory([host.to_dict() for host in hosts])
         return hosts if not len(hosts) == 1 else hosts[0]
```

tpapaioa:
Fixes #92

### DIFF
--- a/broker/broker.py
+++ b/broker/broker.py
@@ -100,10 +100,8 @@ class VMBroker:
             return result
 
     @mp_decorator
-    def _checkout(self, connect=False):
+    def _checkout(self):
         """checkout one or more VMs
-
-        :param connect: Boolean whether to establish host ssh connection
 
         :return: List of Host objects
         """
@@ -112,10 +110,8 @@ class VMBroker:
             provider, method = PROVIDER_ACTIONS[action]
             logger.info(f"Using provider {provider.__name__} to checkout")
             host = self._act(provider, method, checkout=True)
-            logger.debug(f"host={host} connect={connect}")
+            logger.debug(f"host={host}")
             if host:
-                if connect:
-                    host.connect()
                 hosts.append(host)
                 logger.info(f"{host.__class__.__name__}: {host.hostname}")
         return hosts
@@ -127,7 +123,10 @@ class VMBroker:
 
         :return: Host obj or list of Host objects
         """
-        hosts = self._checkout(connect=connect)
+        hosts = self._checkout()
+        if connect:
+            for host in hosts:
+                host.connect()
         self._hosts.extend(hosts)
         helpers.update_inventory([host.to_dict() for host in hosts])
         return hosts if not len(hosts) == 1 else hosts[0]


JacobCallahan (approved):
ACK, thanks!

PR #98, opened 2021-04-13
Add tower_inventory to _broker_args

JacobCallahan:
Explicitly check for and add tower_inventory to _broker_args.
This will help with things like duplicate and checkin.

Also added a yaml representer for awxkit PseudoNamespace objects.


mshriver (approved):

mshriver (changes requested):
~Looked good on inspection, but the yaml representer is writing a bad tag into the inventory yaml file:~

Edit: This was due to a bad venv.

mshriver (approved):

PR #102, opened 2021-04-23
Added some enhancements to the default Host class

JacobCallahan:
It is now possible to define setup and teardown behavior
HostErrors now how their own exception class

### DIFF
--- a/broker/broker.py
+++ b/broker/broker.py
@@ -14,7 +14,7 @@ PROVIDER_ACTIONS = {
     "job_template": (AnsibleTower, "execute"),
     "template": (AnsibleTower, None),  # needed for list-templates
     "test_action": (TestProvider, "test_action"),
-    "inventory": (AnsibleTower, None)
+    "inventory": (AnsibleTower, None),
 }
 
 
@@ -284,10 +284,21 @@ class VMBroker:
             hosts = self.checkout(connect=True)
             if not hosts:
                 raise Exception("No hosts created during checkout")
+            [host.setup() for host in hosts]
             return hosts
         except Exception as err:
             self.checkin()
             raise err
 
     def __exit__(self, exc_type, exc_value, exc_traceback):
+        last_exception = None
+        for host in self._hosts:
+            try:
+                host.teardown()
+            except Exception as err:
+                last_exception = exceptions.HostError(
+                    host, f"error during teardown:\n{err}"
+                )
         self.checkin()
+        if last_exception:
+            raise last_exception
--- a/broker/exceptions.py
+++ b/broker/exceptions.py
@@ -34,3 +34,11 @@ class ConfigurationError(BrokerError):
 
 class NotImplementedError(BrokerError):
     error_code = 9
+
+
+class HostError(BrokerError):
+    error_code = 10
+
+    def __init__(self, host, message):
+        self.message = f"{host.hostname or host.name}: {message}"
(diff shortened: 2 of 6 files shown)


broker/exceptions.py
```diff
@@ -34,3 +34,11 @@ class ConfigurationError(BrokerError):
 
 class NotImplementedError(BrokerError):
     error_code = 9
+
+
+class HostError(BrokerError):
+    error_code = 10
+
+    def __init__(self, host, message):
+        self.message = f"{host.hostname or host.name}: {message}"
```
tstrych, line 43: Why the `or` is there ? When the `host.hostname` can be None or empty ?

broker/session.py
```diff
@@ -57,10 +57,11 @@ def _read(self, channel):
             stderr=channel.read_stderr(),
         )
 
-    def run(self, command):
+    def run(self, command, timeout=0):
         """run a command on the host and return the results"""
+        self.session.set_timeout(timeout)
```
tstrych, line 62: where is the set_timeout defined ? 
I found only [settimeout](https://docs.python.org/3/library/socket.html#socket.socket.settimeout)

broker/session.py
```diff
@@ -57,10 +57,11 @@ def _read(self, channel):
...
         channel = self.session.open_session()
-        channel.execute(command)
+        channel.execute(command, )
```
tstrych, line 64: nitpick, `(command)`
AFAIK, in one liners we do not use commas at the right ?

tstrych (approved):
ACK, small comments, please answer my question :+1:

broker/exceptions.py
JacobCallahan: Exactly. It shouldn't happen, but just in case, we have a fallback

broker/session.py
JacobCallahan: self.session is an ssh2-python session object, try checking the docs for that library and verify I'm not mistaken

broker/session.py
JacobCallahan: This is a black auto-format change. Generally there isn't much of a point to change these back since black will attempt to format them again.

JacobCallahan:
@tstrych thanks for checking this out! I believe I addressed all your questions

broker/session.py
tstrych: yes you are right, I looked at wrong repo :) [ssh2-python set_timeout](https://ssh2-python.readthedocs.io/en/latest/session.html?highlight=set_timeout#ssh2.session.Session.set_timeout)

broker/broker.py
```diff
@@ -284,10 +284,21 @@ def __enter__(self):
...
             self.checkin()
             raise err
 
     def __exit__(self, exc_type, exc_value, exc_traceback):
+        last_exception = None
+        for host in self._hosts:
+            try:
+                host.teardown()
+            except Exception as err:
+                last_exception = exceptions.HostError(
```
mshriver, line 299: I think we should `logger.exception` here so that its at least recorded. If there are exceptions on multiple hosts, the user will only be aware of the last exception to occur during teardown.

JacobCallahan: automatically done when init is called :)

mshriver: AH `BrokerError.__init__` emits, sweet.

PR #103, opened 2021-04-26
Turn host.session into a property

mshriver:
If connect wasn't called, session can be referenced when it was None. Callers should be able to create a host instance and use its `execute` method without being forced to connect first. Broker should be responsible for handling that session connection.

This will replace session whenever `connect` is called.

Since connect is handled automatically along with the session property, there is no longer a need to track connected state through pickling, so `__setstate__` has been removed.

Example, where `host` was created with `VMBroker.from_inventory()`:

```
In [7]: host.execute('hostname')
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-7-0a254b67faf4> in <module>
----> 1 host.execute('hostname')

~/repos/robottelo/.robottelo/lib64/python3.8/site-packages/broker/hosts.py in execute(self, command, timeout)
     49         timeout = timeout or self.default_timeout
     50         logger.debug(f"{self.hostname} executing command: {command}")
---> 51         res = self.session.run(command, timeout=timeout)
     52         logger.debug(f"{self.hostname} command result:\n{res}")
     53         return res

AttributeError: 'NoneType' object has no attribute 'execute'
```

This isn't a reasonable failure mode for broker's host.  The reference to session should create it if it doesn't exist yet.

With this change, `host.execute` can be called without having to explicitly call `host.connect` first.

With my example system being powered off, this results in a TimeoutError, as I would expect, instead of an AttributeError.

```
In [5]: host.run('hostname')
---------------------------------------------------------------------------
TimeoutError                              Traceback (most recent call last)
<ipython-input-5-0a254b67faf4> in <module>
----> 1 host.run('hostname')

~/repos/broker/broker/hosts.py in execute(self, command, timeout)
     70         timeout = timeout or self.default_timeout
     71         logger.debug(f"{self.hostname} executing command: {command}")
---> 72         res = self.session.run(command, timeout=timeout)
     73         logger.debug(f"{self.hostname} command result:\n{res}")
     74         return res

~/repos/broker/broker/hosts.py in session(self)
     24     def session(self):
     25         if not self._session:
---> 26             self.connect()
     27         return self._session
     28 

~/repos/broker/broker/hosts.py in connect(self, username, password)
     45         if self._session is not None:
     46             self.close()
---> 47         self._session = session.Session(
     48             hostname=self.hostname, username=username, password=password
     49         )

~/repos/broker/broker/session.py in __init__(self, **kwargs)
     36         user = kwargs.get("username", "root")
     37         sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
---> 38         sock.connect((host, kwargs.get("port", 22)))
     39         self.session = ssh2_Session()
     40         self.session.handshake(sock)

TimeoutError: [Errno 110] Connection timed out

```

### DIFF
--- a/broker/hosts.py
+++ b/broker/hosts.py
@@ -17,32 +17,38 @@ class Host:
             self.hostname = hostname
             self.name = name
         self.username = kwargs.get("username", settings.HOST_USERNAME)
-        self.password = kwargs.get("pwassword", settings.HOST_PASSWORD)
-        self.session = None
+        self.password = kwargs.get("password", settings.HOST_PASSWORD)
+        self._session = None
+
+    def __del__(self):
+        """Try to close the connection on garbage collection of the host instance"""
+        self.close()
+        # object.__del__ DNE, so I don't have to call it here.
+        # If host inherits from a different class with __del__, it should get called through super
+
+    @property
+    def session(self):
+        if not isinstance(self._session, session.Session):
+            self.connect()
+        return self._session
 
     def __getstate__(self):
         """If a session is active, remove it for pickle compatability"""
-        if self.session:
-            self.__connected = True
-            self.session = None
+        self.close()
         return self.__dict__
-
-    def __setstate__(self, pickle_dict):
-        """If a session was active pre-pickle, reconnect post-pickle"""
-        self.__dict__ = pickle_dict
-        if self.__dict__.pop('__connected', False):
-            self.connect()
 
     def connect(self, username=None, password=None):
         username = username or self.username
         password = password or self.password
-        self.session = session.Session(
+        self.close()
+        self._session = session.Session(
             hostname=self.hostname, username=username, password=password
         )
 
     def close(self):
-        if isinstance(self.session, session.Session):
-            self.session.session.disconnect()
+        if isinstance(self._session, session.Session):
+            self._session.session.disconnect()
+            self._session = None
 
     def release(self):
         raise NotImplementedError("release has not been implemented for this provider")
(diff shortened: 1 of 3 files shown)


JacobCallahan (changes requested):
Can you use close in getstate? That is cleaner than the current implementation

JacobCallahan:
With session becoming a property, we no longer need to call or check for `connect` in the host, context managers, and other places relating to host construction.
Would you mind removing those pieces as part of this PR as well?

mshriver:
@JacobCallahan Absolutely, wanted to hear some buy-in on changing the pattern here for the session handling first. Likely a few small changes that could happen in robottelo as well.

JacobCallahan:
@mshriver absolutely! Turning this into a property is going to be VERY nice!

broker/hosts.py
```diff
@@ -17,32 +17,38 @@ def __init__(self, hostname, name=None, from_dict=False, **kwargs):
             self.hostname = hostname
             self.name = name
         self.username = kwargs.get("username", settings.HOST_USERNAME)
-        self.password = kwargs.get("pwassword", settings.HOST_PASSWORD)
-        self.session = None
+        self.password = kwargs.get("password", settings.HOST_PASSWORD)
```
mshriver, line 20: Note the typo that was here is getting fixed.

JacobCallahan: hahaha nice!!

JacobCallahan (approved):

PR #107, opened 2021-05-10
Added remote_copy to the Host's session object

JacobCallahan:
This enables direct file copying from one broker host object to another!

### DIFF
--- a/broker/providers/ansible_tower.py
+++ b/broker/providers/ansible_tower.py
@@ -238,25 +238,33 @@ class AnsibleTower(Provider):
         expire_time = self._get_expire_date(host.id)
         if expire_time:
             host_info["expire_time"] = expire_time
-        if "last_job" in host.related:
-            last_job = host.get_related("last_job")
-            job_vars = json.loads(last_job.extra_vars)
-            host_info["_broker_args"].update(
-                {
-                    arg: val
-                    for arg, val in job_vars.items()
-                    if val and isinstance(val, str)
-                }
-            )
-            host_info["_broker_args"]["tower_inventory"] = last_job.inventory
-            try:
-                host_info["_broker_args"]["workflow"] = host.get_related(
-                    "last_job"
-                ).summary_fields.source_workflow_job.name
-            except:
-                logger.warning(
-                    f"Unable to determine workflow for {host_info['hostname']}"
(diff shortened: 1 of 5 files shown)


broker/providers/ansible_tower.py
```diff
@@ -238,25 +238,33 @@ def _compile_host_info(self, host):
...
-            except:
-                logger.warning(
-                    f"Unable to determine workflow for {host_info['hostname']}"
-                )
+        try:
+            create_job = self.v2.jobs.get(id=host.get_related("job_events").results[0].job)
+            create_job = create_job.results[0].get_related("source_workflow_job")
+            host_info["_broker_args"]["workflow"] = create_job.name
+        except IndexError:
+            if "last_job" in host.related:
```
JacobCallahan, line 246: At this point, we could really just warn the user that the data is likely unreliable and won't include any information. However, we'd either have to put in a mechanism where the host isn't included in the inventory or let the _broker_args be mostly blank.

tstrych:
little late to party :partying_face: but never mind :smile: 
never is too late
thanks jake for helping me with setup :smiling_face_with_three_hearts: 

I tried my usual sync I use
`broker inventory --sync AnsibleTower --details` 
works as expected.

I did not try the exception, because I don't know how @JacobCallahan, i guess when the job fail there is no result 

how can I make the job fail ?

I tried the copying based on picture you send us 
it worked :+1: 

but have some questions and comments :1234: 

firstly can we get some error, like file doesn't exist ? 
instead of `SFTPProtocolError:`

and I guess something with file existence, appending doesn't work as I would expect.

My example:
```
<host> echo "test text" > test
<ipython> host.session.remote_copy('test',sat)
<sat> cat test
test text


<host> # i removed the text from the test  file
<host> cat test
test
<ipython> host.session.remote_copy('test',sat)
<sat> cat test
test
text
```

Expected result
file already exists or
``` 
<sat> cat test 
test
```

JacobCallahan:
@tstrych thanks for testing!

Regarding error handling and file existence checking, those are largely beyond the scope of this Session class. This class is solely a very thin wrapper around ssh2-python. Errors and existence tests should be handled by whoever is using broker's Host class to perform actions.

PR #108, opened 2021-05-24
Adding error message when no flags are specified

Griffin-Sullivan:
Added an error message: "Could not determine an appropriate provider" when the user does not specify options to certain commands. Ex: `broker execute` or `broker checkout` with no workflow or nick.

### DIFF
--- a/broker/broker.py
+++ b/broker/broker.py
@@ -156,10 +156,10 @@ class VMBroker:
                 if key in kwargs:
                     self._provider_actions[key] = action
         self._kwargs.update(kwargs)
-        for action, arg in self._provider_actions.items():
-            provider, method = PROVIDER_ACTIONS[action]
         if not self._provider_actions:
             raise self.BrokerError("Could not determine an appropriate provider")
+        for action, arg in self._provider_actions.items():
+            provider, method = PROVIDER_ACTIONS[action]
         logger.info(f"Using provider {provider.__name__} for execution")
         return self._act(provider, method)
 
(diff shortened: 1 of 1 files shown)


Griffin-Sullivan:
Just realized I did not test this enough. Going to update the commit with a fix.

Griffin-Sullivan:
Ok from what I tested, this should work. Also, I'm not sure what I did to make my commits so weird, so if someone could help resolve that, it would be much appreciated.

broker/broker.py
```diff
@@ -92,9 +92,14 @@ def __init__(self, **kwargs):
...
         # determine the provider actions based on kwarg parameters
         self._provider_actions = {}
+        no_provider = False
         for key, action in PROVIDER_ACTIONS.items():
             if key in kwargs:
                 self._provider_actions[key] = action
+            else:
+                no_provider = True
+        if no_provider:
+            logger.error("Could not determine an appropriate provider")
```
JacobCallahan, line 102: We can simplify further!
```suggestion
        for key, action in PROVIDER_ACTIONS.items():
            if key in kwargs:
                self._provider_actions[key] = action
        if not self._provider_actions:
            logger.error("Could not determine an appropriate provider")
```

JacobCallahan (changes requested):

broker/broker.py
Griffin-Sullivan: Oh I see. I didn't think to check that `_provider_actions` was null that way.

JacobCallahan (approved):

PR #116, opened 2021-07-01
Remove travis, Add workflow for pypi publishing

mshriver:
Some small updates around the core goal of moving pypi publishing to github workflows.

Adjust the run `on` specifications for all workflows.

Trying out https://github.com/nektos/act to test the workflows. Secrets for container push will only be available on the main fork.


requirements.txt
```diff
@@ -0,0 +1 @@
+.
```
JacobCallahan, line 1: what is this needed for?

mshriver: This was from Jaro's commits, but its a nice little convenience thing.  If you do the common `pip install -Ur requirements.txt`, this will cause it to pickup the install requirements defined in setup.py.

JacobCallahan: ehh... I can't say I'm a fan of the pattern

mshriver: ack, I can remove.

JacobCallahan: sweet. i don't want to promote the requirements.txt pattern in this repo

jaryn: @JacobCallahan Can you please explain why did you not accept this change? Thanks.

PR #117, opened 2021-07-01
Improved concurrent checkin resilience

JacobCallahan:
Also increased AnsibleTower page size for workflows and job templates

PR #119, opened 2021-07-02
Improved concurrent checkin resilience

JacobCallahan:
Also increased AnsibleTower page size for workflows and job templates.
Using the _prov_inst also made it so we only auth once per CLI host checkin!
A little bit of black formatting happened as well..

This is an alternative PR to #117

### DIFF
--- a/broker/broker.py
+++ b/broker/broker.py
@@ -100,11 +100,13 @@ class VMBroker:
     def _act(self, provider, method, checkout=False):
         """Perform a general action against a provider's method"""
         provider_inst = provider(**self._kwargs)
-        helpers.emit({
-            "provider": provider_inst.__class__.__name__,
-            "action": method,
-            "arguments": self._kwargs
-        })
+        helpers.emit(
+            {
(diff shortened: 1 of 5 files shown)


broker/broker.py
```diff
@@ -195,23 +195,32 @@ def checkin(self, sequential=False, host=None):
...
             completed_checkins = as_completed(
                 # reversing over a copy of the list to avoid skipping
                 workers.submit(self._checkin, _host)
                 for _host in hosts[::-1]
             )
             for completed in completed_checkins:
                 _host = completed.result()
-                logger.debug(f'Completed checkin process for {_host.hostname or _host.name}')
+                self._hosts = [
+                    h for h in self._hosts if not (h.to_dict() == _host.to_dict())
```
mshriver, line 219: Consider implementing `Host.__eq__` so we can directly compare host instances.

mshriver (approved):

broker/broker.py
JacobCallahan: that's not a bad idea

PR #123, opened 2021-07-16
Check for empty hosts in VMBroker.checkin

mshriver:
I found that checkin was getting called against an instance that had no
hosts, triggering a ValueError because 0 was getting passed for
max_workers to ProcessPoolExecutor

Changed the context manager to not capture and try to checkin when no
hosts were created.

Checked that hosts isn't empty in VMBroker.checkin

Added unit test that reproduced the error

### DIFF
--- a/tests/test_broker.py
+++ b/tests/test_broker.py
@@ -44,6 +44,13 @@ def test_broker_e2e():
     assert len(broker_inst._hosts) == 0
 
 
+def test_broker_empty_checkin():
+    """Try to checkin with no hosts on the instance"""
+    broker_inst = broker.VMBroker(nick="test_nick")
+    assert not broker_inst._hosts
+    broker_inst.checkin()
+
+
 def test_mp_checkout():
     """Test that broker can checkout multiple hosts using multiprocessing"""
     VM_COUNT=50  # This is intentionaly made high to catch run condition that
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,6 +1,12 @@
 History
 =======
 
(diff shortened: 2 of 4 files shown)


JacobCallahan (approved):

latran (approved):
LGTM

tests/test_broker.py
```diff
@@ -44,6 +44,13 @@ def test_broker_e2e():
     assert len(broker_inst._hosts) == 0
 
 
+def test_broker_empty_checkin():
+    """Try to checkin with no hosts on the instance"""
+    broker_inst = broker.VMBroker(nick="test_nick")
+    assert not len(broker_inst._hosts)
```
tstrych, line 50: I found [here](https://stackoverflow.com/questions/53513/how-do-i-check-if-a-list-is-empty) people discussing why it's recommended, more pythonic way to use:
`not broker_inst._hosts` instead of `not len(broker_inst._hosts)`

tstrych (approved):
Thanks mike for looking at results. ACK.
Please @mshriver provide the test results.

tests/test_broker.py
tstrych: But at the end of the day it doesn't matter.

JacobCallahan: it's about twice as fast as well :)

mshriver: Good point @tstrych, I think I was rushing a bit, because this is a pattern I regularly recommend and use in robottelo :laughing:

JacobCallahan (approved):

PR #128, opened 2021-08-10
Move to setup.cfg, use setuptools-scm for versioning

mshriver:
This commit implies we will begin to use git tags to control versioning in broker, instead of a static version number in setup.py.

We'll still manually update HISTORY.md with contextual information about tagged versions.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,43 @@
-[bdist_wheel]
-universal = 1
+[metadata]
+name = broker
+description = The infrastructure middleman.
+long_description = file: README.md, HISTORY.md
+long_description_content_type = text/markdown
+author = Jacob J Callahan
+author_email = jacob.callahan05@gmail.com
+url = https://github.com/SatelliteQE/broker
+license = GNU General Public License v3
+keywords = broker, AnsibleTower
+classifiers =
+    Development Status :: 4 - Beta
+    Intended Audience :: Developers
+    License :: OSI Approved :: GNU General Public License v3 (GPLv3)
+    Natural Language :: English
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
+
+[options]
+install_requires =
+    awxkit
+    click
+    dynaconf>=3.1.0
+    logzero
+    pyyaml
+    setuptools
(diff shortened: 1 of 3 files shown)


JacobCallahan (approved):

setup.py
```diff
@@ -1,48 +1,7 @@
...
-    classifiers=[
-        "Development Status :: 4 - Beta",
-        "Intended Audience :: Developers",
-        "License :: OSI Approved :: GNU General Public License v3 (GPLv3)",
-        "Natural Language :: English",
-        "Programming Language :: Python :: 3",
-        "Programming Language :: Python :: 3.8",
-        "Programming Language :: Python :: 3.9",
-    ],
+    use_scm_version=True,
```
mshriver, line 6: This argument here is the setuptools-scm integration that will read version dynamically from SCM tags (and other fields, but tags is the intended use).

setup.cfg
```diff
@@ -1,2 +1,43 @@
...
+[metadata]
+name = broker
+description = The infrastructure middleman.
+long_description = file: README.md, HISTORY.md
+long_description_content_type = text/markdown
+author = Jacob J Callahan
+author_email = jacob.callahan05@gmail.com
+url = https://github.com/SatelliteQE/broker
+license = GNU General Public License v3
+keywords = broker, AnsibleTower
```
mshriver, line 10: Adding the AnsibleTower keyword here

tstrych: just wanted to ask you about that :)

setup.cfg
```diff
@@ -1,2 +1,43 @@
-[bdist_wheel]
-universal = 1
```
tstrych, line 2: why bdist_wheel --universal is no longer needed ?

tstrych (approved):

tstrych (approved):

setup.cfg
mshriver: universal wheels are for py2+py3 support, we're on py 3.8+ now here and no longer need universal wheels.

https://packaging.python.org/guides/distributing-packages-using-setuptools/#wheels

setup.cfg
mshriver: Just increasing visibility/discoverabilty

PR #129, opened 2021-08-10
Better job error handling for AnsibleTower

JacobCallahan:
This change allows broker to also handle jobs that error out instead
of simply failing. It also provides a generic error message if none
could be found.

### DIFF
--- a/broker/commands.py
+++ b/broker/commands.py
@@ -45,6 +45,8 @@ def populate_providers(click_group):
         --workflows      Get available workflows
         --workflow TEXT  Get information about a workflow
         --help           Show this message and exit.
+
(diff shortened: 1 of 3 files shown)


broker/commands.py
```diff
@@ -45,6 +45,8 @@ def populate_providers(click_group):
         --workflows      Get available workflows
         --workflow TEXT  Get information about a workflow
         --help           Show this message and exit.
+
+    Note: This currently only works for the default instance for each provider
```
mshriver, line 49: :+1:

JacobCallahan: Thought i'd sneak that in there.

mshriver (approved):

PR #137, opened 2021-09-10
Add interactive shell

JacobCallahan:
This change adds a method to the session object which allows it to spawn
an interactive shell in a new channel.
Also made some small tweaks to fix an issue with closing sessions.

### DIFF
--- a/broker/session.py
+++ b/broker/session.py
@@ -45,7 +45,8 @@ class Session:
         else:
             raise AuthException("No password or key file provided.")
 
-    def _read(self, channel):
+    @staticmethod
+    def _read(channel):
         """read the contents of a channel"""
         size, data = channel.read()
         results = ""
@@ -62,10 +63,17 @@ class Session:
         """run a command on the host and return the results"""
         self.session.set_timeout(timeout)
         channel = self.session.open_session()
-        channel.execute(command, )
+        channel.execute(
+            command,
+        )
         results = self._read(channel)
         channel.close()
         return results
+
+    def shell(self, pty=False):
+        """Create and return an interactive shell instance"""
+        channel = self.session.open_session()
+        return InteractiveShell(channel, pty)
 
     def sftp_read(self, source, destination=None):
         """read a remote file into a local destination"""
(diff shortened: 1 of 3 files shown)


broker/session.py
```diff
@@ -126,3 +134,47 @@ def __enter__(self):
...
+    def send(self, cmd):
+        """Send a command to the channel, ensuring a newline character"""
+        if not cmd.endswith("\n"):
+            cmd += "\n"
+        self._chan.write(cmd)
+
+    def stdout(self):
+        """read the contents of a channel's stdout"""
+        if not self._chan.eof():
+            _, data = self._chan.read(65535)
```
tstrych, line 172: do we like magic number constants directly in our code ?

broker/session.py
```diff
@@ -126,3 +134,47 @@ def __enter__(self):
 
     def __exit__(self, *args):
         self.session.disconnect()
+
+
+class InteractiveShell:
+    """A helper class that provides an interactive shell interface"""
```
tstrych, line 140: the provided explanation is not telling  much more than the name of the class itself.

please provide an explanation when this can be used or why it should be used, maybe you can add some usage examples

tstrych (commented):
Some comments. Not necessarily  blocking but I think nice to have.

broker/session.py
```diff
@@ -62,11 +63,18 @@ def run(self, command, timeout=0):
         """run a command on the host and return the results"""
         self.session.set_timeout(timeout)
         channel = self.session.open_session()
-        channel.execute(command, )
+        channel.execute(
+            command,
+        )
```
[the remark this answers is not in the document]
mshriver: I'm guessing this is `Black` formatting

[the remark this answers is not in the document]
JacobCallahan: likely overzealous black

broker/session.py
JacobCallahan: no and this hopefully won't stay in the long run. it is a much higher value than the lower number of bits read by default. ideally this whole method would be a lot more read safe for instances where there is no EOF

broker/session.py
JacobCallahan: I can add one brief, and safe, example here. Overall, I'd love to have a larger example doc(s), but for now...

mshriver (approved):

broker/session.py
JacobCallahan: @tstrych addressed in latest commit

tstrych: thanks :partying_face:

tstrych (approved):

PR #138, opened 2021-09-17
Add back in getattr call to Host methods

JacobCallahan:
These are still needed since self._session is stripped during the
pickling process.


broker/hosts.py
```diff
@@ -29,7 +29,7 @@ def __del__(self):
 
     @property
     def session(self):
-        if not isinstance(self._session, Session):
+        if not isinstance(getattr(self, "_session", None), Session):
```
mshriver, line 32:
```suggestion
        # This attribute may be missing after pickling
        if not isinstance(getattr(self, "_session", None), Session):
```

broker/hosts.py
```diff
@@ -56,7 +56,7 @@ def connect(self, username=None, password=None):
         )
 
     def close(self):
-        if isinstance(self._session, Session):
+        if isinstance(getattr(self, "_session", None), Session):
```
mshriver, line 59:
```suggestion
        # This attribute may be missing after pickling
        if isinstance(getattr(self, "_session", None), Session):
```

broker/hosts.py
JacobCallahan: done

broker/hosts.py
JacobCallahan: done

mshriver (approved):

PR #139, opened 2021-09-28
Fixing things up in preparation for larger changes

### LINKED ISSUE #133: Extending the life time of multiple VMs takes too much of time.
Extending the VMs is done one by one, waiting for each job to complete sequentially. Much of time can be saved from broker runtime if done concurrently, for example using asyncio.

### LINKED ISSUE #135: File logs in broker contain terminal colouring
This may be problemattic when displaying as plaintext.

JacobCallahan:
+ Extends are now concurrent by default
+ Color codes are stripped from log files
+ Broker now respects its own debug setting
+ Session command timeouts are now flexible
Fixes #133
Fixes #135

### DIFF
--- a/broker/helpers.py
+++ b/broker/helpers.py
@@ -175,7 +175,7 @@ def resolve_file_args(broker_args):
         if isinstance(val, Path) or (
             isinstance(val, str) and val[-4:] in ("json", "yaml", ".yml")
         ):
-            if (data := load_file(val)) :
+            if (data := load_file(val)):
                 if key == "args_file":
                     if isinstance(data, dict):
                         final_args.update(data)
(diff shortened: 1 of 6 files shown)


mshriver (approved):

broker/helpers.py
```diff
@@ -361,4 +361,26 @@ def handle_keyboardinterrupt(*args):
...
+            unit = "h"
+        if unit == "h":
+            timeout *= 60
+            unit = "m"
+        if unit == "m":
+            timeout *= 60
+            unit = "s"
+        if unit == "s":
+            timeout *= 1000
+    return timeout if isinstance(timeout, int) else 0
```
tstrych, line 386: there was no and still is no new line at the end of the file, this is first time I noticed that github is showing such a thing

tstrych (approved):
ACK

broker/helpers.py
JacobCallahan: oh weird. good catch!

JacobCallahan: and now fixed

PR #140, opened 2021-10-04
Added retries to socket.connect

JacobCallahan:
Also added some better error handling when decoding stdout data

### DIFF
--- a/broker/helpers.py
+++ b/broker/helpers.py
@@ -3,6 +3,7 @@ import json
 import logging
 import os
 import sys
+import time
 from collections import UserDict, namedtuple
 from collections.abc import MutableMapping
 from copy import deepcopy
@@ -175,7 +176,7 @@ def resolve_file_args(broker_args):
         if isinstance(val, Path) or (
             isinstance(val, str) and val[-4:] in ("json", "yaml", ".yml")
         ):
-            if (data := load_file(val)):
+            if (data := load_file(val)) :
                 if key == "args_file":
                     if isinstance(data, dict):
                         final_args.update(data)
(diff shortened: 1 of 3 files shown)


broker/helpers.py
```diff
@@ -384,3 +386,21 @@ def translate_timeout(timeout):
         if unit == "s":
             timeout *= 1000
     return timeout if isinstance(timeout, int) else 0
+
+
+def simple_retry(cmd, cmd_args=None, cmd_kwargs=None, max_timeout=60, _cur_timeout=1):
+    """Re(Try) a function given its args and kwargs up until a max timeout"""
+    cmd_args = cmd_args if cmd_args else []
+    cmd_kwargs = cmd_kwargs if cmd_kwargs else {}
+    try:
```
mshriver, line 395: Why not use wait_for here?

broker/session.py
```diff
@@ -52,7 +53,10 @@ def _read(channel):
         size, data = channel.read()
         results = ""
         while size > 0:
-            results += data.decode("utf-8")
+            try:
+                results += data.decode("utf-8")
+            except UnicodeDecodeError as err:
+                logger.error(f"Skipping data chunk due to {err}\nReceived: {data}")
```
mshriver, line 59: This becomes invisible to the caller at runtime since no exception is raised.  We could easily be asserting against stdout that is missing a chunk, and pass those assertions while never knowing this exception occurred.

broker/helpers.py
JacobCallahan: i don't want an additional requirement for such a simple pattern

broker/session.py
JacobCallahan: It does indeed, but might be the best current option.

PR #143, opened 2021-11-05
Greatly speed up Session's sftp_write

JacobCallahan:
This change no longer initiates a sftp write for each line of an
uploaded file. Instead, it passes the entire file to sftp's write and
allows libssh2 to more efficiently control buffering.


mshriver (approved):

tstrych (approved):

tstrych:
CodeQL Analysis (3.10) failed - `undefined symbol: _PyGen_Send`

PR #144, opened 2021-11-11
Added a locking mechanism to actions that modify files

JacobCallahan:
This should help us mitigate the occasional race conditions when
multiple instances of broker are attempting to update a single local
inventory file.

### DIFF
--- a/broker/helpers.py
+++ b/broker/helpers.py
@@ -177,7 +177,7 @@ def resolve_file_args(broker_args):
         if isinstance(val, Path) or (
             isinstance(val, str) and val[-4:] in ("json", "yaml", ".yml")
         ):
-            if (data := load_file(val)) :
+            if data := load_file(val):
                 if key == "args_file":
                     if isinstance(data, dict):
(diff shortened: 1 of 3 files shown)


broker/helpers.py
```diff
@@ -223,20 +223,21 @@ def update_inventory(add=None, remove=None):
...
+            inventory_file.unlink()
 
-    if remove:
-        for host in inv_data[::-1]:
-            if host["hostname"] in remove or host["name"] in remove:
-                inv_data.remove(host)
-    if add:
-        inv_data.extend(add)
+        if remove:
+            for host in inv_data[::-1]:
```
mshriver, line 232: Do we want to lock for the whole time we're iterating over inventory data?

mshriver (approved):

broker/helpers.py
JacobCallahan: yep indeed. otherwise, we could run into situations where another process reads the inventory we're about to modify then overwrite our changes.

PR #145, opened 2021-11-30
Adding a connection timeout option

JacobCallahan:
This timeout is both available as a setting as well as when creating a Session instance.

### DIFF
--- a/broker/providers/ansible_tower.py
+++ b/broker/providers/ansible_tower.py
@@ -271,7 +271,7 @@ class AnsibleTower(Provider):
                                         "reason": ev.event_data["res"]["msg"],
                                     }
                                     for ev in failed_events
-                                    if ev.event_data.get("res")
+                                    if ev.event_data.get("res", {}).get("msg")
                                 ]
                             )
         if not failure_messages:
--- a/broker/session.py
+++ b/broker/session.py
@@ -37,6 +37,7 @@ class Session:
         host = kwargs.get("hostname", "localhost")
         user = kwargs.get("username", "root")
         sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sock.settimeout(kwargs.get("timeout"))
         simple_retry(sock.connect, [(host, kwargs.get("port", 22))])
         self.session = ssh2_Session()
         self.session.handshake(sock)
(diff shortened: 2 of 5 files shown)


broker/session.py
```diff
@@ -37,6 +37,8 @@ def __init__(self, **kwargs):
         host = kwargs.get("hostname", "localhost")
         user = kwargs.get("username", "root")
         sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        if timeout := kwargs.get("timeout"):
+            sock.settimeout(timeout)
```
jyejare, line 41: We will always have a timeout value since we are setting the default from `settings.yaml` here and unless someone explicitly set it to None from the `Host` object or `connect{)` def.

So simply we can:
```suggestion
        sock.settimeout(kwargs.get("timeout"))
```

broker/settings.py
```diff
@@ -18,6 +18,7 @@
 validators = [
     Validator("HOST_USERNAME", default="root"),
     Validator("HOST_PASSWORD", must_exist=True),
+    Validator("HOST_CONNECTION_TIMEOUT", default=1),
```
jyejare, line 21: Does the default `1` here set for testing purpose, if yes `thats cool` and if not:
1. does `1` denotes unlimited timeout? which I dont believe so
2. OR why it's so less?

jyejare (commented):
ACK with some pending comments / questions.

Reviewed Draft MR, I can revisit though is anything changes!

broker/session.py
JacobCallahan: Your suggestion works in general since `settimeout` can accept a value of `None`

broker/settings.py
JacobCallahan: Yes, 1 was a placeholder value and a numeric value is in seconds

broker/providers/ansible_tower.py
```diff
@@ -268,7 +268,7 @@ def _get_failure_messages(self, workflow):
                                     {
                                         "job": child_obj.name,
                                         "task": ev.event_data["play"],
-                                        "reason": ev.event_data["res"]["msg"],
+                                        "reason": ev.event_data["res"].get("msg"),
```
mshriver, line 271: Consider using `.get('msg', 'No failure message returned by awx')`

mshriver (approved):

PR #151, opened 2022-03-31
Introducing the Container provider

JacobCallahan:
This pr introduces the container provider.
The first supported bind is Docker, though a Podman bind is in progress.
A number of other supporting concepts have also been introduced, most
notably are "binds". These are small abstraction layers around an
external API that ease the integration into Broker.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,7 @@ __pycache__/
 # Distribution / packaging
 .Python
 env/
-venv*/
+ven*/
 build/
 develop-eggs/
 dist/
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ The infrastrucure middleman
 
 
 # Description
-Broker is a tool designed to provide a common interface between one or many services that provision virtual machines. It is an abstraction layer that allows you to ignore (most) of the implementation details and just get what you need.
+Broker is a tool designed to provide a common interface between one or many services that provision virtual machines or containers. It is an abstraction layer that allows you to ignore most of the implementation details and just get what you need.
 
 # Installation
 ```
@@ -17,6 +17,14 @@ cp broker_settings.yaml.example broker_settings.yaml
 ```
 Then edit the broker_settings.yaml file
 
+If you are using the Container provider, then install the extra dependency based on your container runtime of choice.
+```
+pip install broker[podman]
+or
+pip install broker[docker]
+```
+These may not work correctly in non-bash environments.
+
 Broker can also be ran outside of its base directory. In order to do so, specify the directory broker's files are in with the
 `BROKER_DIRECTORY` envronment variable.
 ```BROKER_DIRECTORY=/home/jake/Programming/broker/ broker inventory```
@@ -24,18 +32,18 @@ Broker can also be ran outside of its base directory. In order to do so, specify
 # Configuration
 The broker_settings.yaml file is used, through DynaConf, to set configuration values for broker's interaction with its 'providers'.
 
-DynaConf integration provides support for setting environment variables to override any settings from the yaml file.
+DynaConf integration provides support for setting environment variables to override any settings from broker's config file.
 
-An environment variable override would take the form of: `DYNACONF_AnsibleTower__base_url="https://my.ansibletower.instance.com"`. Note the use of double underscores to model nested maps in yaml.
+An environment variable override would take the form of: `BROKER_AnsibleTower__base_url="https://my.ansibletower.instance.com"`. Note the use of double underscores to model nested maps in yaml.
 
-Broker allows for multiple instances of a provider to be in the configuration file. You can name an instance anything you want, then put instance-specfic settings nested under the instance name. One of your instances must have a setting `default: True`.
+Broker allows for multiple instances of a provider to be in its config file. You can name an instance anything you want, then put instance-specfic settings nested under the instance name. One of your instances must have a setting `default: True`.
 
 For the AnsibleTower provider, authentication can be achieved either through setting a username and password, or through a token (Personal Access Token in Tower).
 
-A username can still be provided when using a token to authenticate. This user will be used for inventory sync (examples below). This may be helpful for AnsibleTower administrators who would like to use their own token to authenticate, but want to set a different user in configuration for checking inventory.
+A username can still be provided when using a token to authenticate. This user will be used for inventory sync (examples below). This might be helpful for AnsibleTower administrators who would like to use their own token to authenticate, but want to set a different user in configuration for checking inventory.
 
-# Usage
-**Checking out a VM**
+# CLI Usage
+**Checking out a VM or container**
 ```
 broker checkout --workflow "workflow-name" --workflow-arg1 something --workflow-arg2 else
 ```
@@ -56,6 +64,7 @@ You can also pass in a file for other arguments, where the contents will become 
 ```
 broker checkout --nick rhel7 --extra tests/data/args_file.yaml
 ```
+**Note:** Check with the provider to determine specific arguments.
 
 **Nicks**
 
@@ -74,9 +83,9 @@ broker duplicate 1 3
 broker duplicate 0 --count 2
 ```
 
-**Listing your VMs**
+**Listing your VMs and containers**
 
-Broker maintains a local inventory of the VMs you've checked out. You can see these with the ```inventory``` command.
+Broker maintains a local inventory of the VMs and containers you've checked out. You can see these with the ```inventory``` command.
 ```
 broker inventory
 ```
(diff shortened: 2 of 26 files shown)


broker/broker.py
```diff
@@ -70,7 +79,7 @@ def mp_split(*args, **kwargs):
         return mp_split
 
 
-class VMBroker:
+class Broker:
```
Griffin-Sullivan, line 82: This change++

JacobCallahan:
Looks like I need to account for differences in unconfigured configs for containers to get the checks to pass on GH

.gitignore
```diff
@@ -9,7 +9,7 @@ __pycache__/
 # Distribution / packaging
 .Python
 env/
-venv*/
+ven*/
```
sambible, line 12: but i call my virtualenv venv :(

sambible: I missed the *

JacobCallahan:
One test failing because AT provider is actually trying to connect, so I need to look into that

README.md
```diff
@@ -6,7 +6,7 @@ The infrastrucure middleman
 
 
 # Description
-Broker is a tool designed to provide a common interface between one or many services that provision virtual machines. It is an abstraction layer that allows you to ignore (most) of the implementation details and just get what you need.
+Broker is a tool designed to provide a common interface between one or many services that provision virtual machines or containers. It is an abstraction layer that allows you to ignore (most) of the implementation details and just get what you need.
```
swadeley, line 9:
```suggestion
Broker is a tool designed to provide a common interface between one or many services that provision virtual machines or containers. It is an abstraction layer that allows you to ignore most of the implementation details and just get what you need.
```
rather leave those brackets out

README.md
```diff
@@ -26,16 +26,16 @@ The broker_settings.yaml file is used, through DynaConf, to set configuration va
 
 DynaConf integration provides support for setting environment variables to override any settings from the yaml file.
```
swadeley, line 27: Re "the yaml file". You use "the" but you have not mentioned a specific file. Is it just general "yaml configuration files"? If so, s/from the yaml file/from yaml configuration files/

README.md
```diff
@@ -26,16 +26,16 @@ The broker_settings.yaml file is used, through DynaConf, to set configuration va
...
 
-An environment variable override would take the form of: `DYNACONF_AnsibleTower__base_url="https://my.ansibletower.instance.com"`. Note the use of double underscores to model nested maps in yaml.
+An environment variable override would take the form of: `BROKER_AnsibleTower__base_url="https://my.ansibletower.instance.com"`. Note the use of double underscores to model nested maps in yaml.
 
 Broker allows for multiple instances of a provider to be in the configuration file. You can name an instance anything you want, then put instance-specfic settings nested under the instance name. One of your instances must have a setting `default: True`.
```
swadeley, line 31: Re: "in the configuration file". What file?

README.md
```diff
@@ -26,16 +26,16 @@ The broker_settings.yaml file is used, through DynaConf, to set configuration va
...
 
 For the AnsibleTower provider, authentication can be achieved either through setting a username and password, or through a token (Personal Access Token in Tower).
 
 A username can still be provided when using a token to authenticate. This user will be used for inventory sync (examples below). This may be helpful for AnsibleTower administrators who would like to use their own token to authenticate, but want to set a different user in configuration for checking inventory.
```
swadeley, line 35: "may" means permission, rather use might or can.

README.md
```diff
@@ -182,3 +185,71 @@ You can also chain multiple filters together by separating them with a comma. Th
...
+result = rhel7_host.execute("rpm -qa")
+assert result.status == 0
+assert "my-package" in result.stdout
+```
+
+
+**Recommended**
+
+The Broker class has a built-in context manager that automatically performs a checkout upon enter and checkin upon exit. It is the recommended way of interacting with Broker for host management.
+In th below two lines of code, a container host is created (pulled if needed/applicable), a broker Host object is constructed, the host object runs a command on the container, output is checked, then the container is checked in.
```
swadeley, line 217:
```suggestion
In the below two lines of code, a container host is created (pulled if needed or applicable), a broker Host object is constructed, the host object runs a command on the container, output is checked, then the container is checked in.
```

README.md
```diff
@@ -182,3 +185,71 @@ You can also chain multiple filters together by separating them with a comma. Th
...
+   ...
+
+with Broker(..., host_classes={'host': MyHost}) as my_host:
+    ...
+```
+
+
+**Setup and Teardown**
+
+Sometimes you may want to define some behavior to occur after a host is checkout out but before Broker gives it to you. Alternatively, you may want to define teardown logic that happens right before a host is checked in.
```
swadeley, line 242: s/you may want/you might want/

broker/broker.py
```diff
@@ -154,6 +160,7 @@ def checkout(self):
                 hosts.remove(host)
         helpers.emit(hosts=[host.to_dict() for host in hosts])
         self._hosts.extend(hosts)
+        helpers.emit(hosts=[host.to_dict() for host in hosts])
```
swadeley, line 163: Did you mean to duplicate this line?

broker/hosts.py
```diff
@@ -50,17 +50,23 @@ def _purify(self):
                 pickle.dumps(obj)
             except (pickle.PicklingError, AttributeError):
                 self.__dict__[key] = None
+            except RecursionError:
+                logger.warning(f"Recursion reached on {obj=}")
```
swadeley, line 54:
```suggestion
                logger.warning(f"Recursion error reached on {obj=}")
```

broker/providers/container.py
```diff
@@ -0,0 +1,278 @@
...
+        name = cont_attrs["name"]
+        logger.debug(f"hostname: {hostname}, name: {name}, host type: host")
+        host_inst = host_classes["host"](
+            **{**kwargs, "hostname": hostname, "name": name}
+        )
+        self._set_attributes(host_inst, broker_args=kwargs, cont_inst=cont_inst)
+        return host_inst
+
+    def nick_help(self, **kwargs):
+        """useful information about container images"""
```
swadeley, line 218:
```suggestion
        """Useful information about container images"""
```

swadeley (approved):
ACK with some typos and one question.

README.md
```diff
@@ -108,9 +109,10 @@ broker extend vmname
 broker extend --all
 ```
 
-**Checking in VMs**
+**Checking in VMs and containers**
 
 You can also return a VM to its provider with the ```checkin``` command.
+Containers checked in this way will be fully deleted regardless of its satus.
```
Griffin-Sullivan, line 115:
```suggestion
Containers checked in this way will be fully deleted regardless of its status.
```

README.md
Griffin-Sullivan, line 242:
```suggestion
Sometimes you may want to define some behavior to occur after a host is checked out but before Broker gives it to you. Alternatively, you may want to define teardown logic that happens right before a host is checked in.
```

README.md
```diff
@@ -90,11 +91,11 @@ broker inventory --sync AnsibleTower:<username>
 ```
 To sync an inventory for a specific instance, use the follow syntax with --sync.
```
Griffin-Sullivan, line 92:
```suggestion
To sync an inventory for a specific instance, use the following syntax with --sync.
```

README.md
```diff
@@ -182,3 +185,71 @@ You can also chain multiple filters together by separating them with a comma. Th
...
+<continued from above>
+class MyHost(Host):
+    ...
+    def setup(self):
+        self.register()
+
+    def teardown(self):
+        self.unregister()
+```
+**note:** One important thing to keep in mind is that Broker will strip any non-pickleable attributes from Host objects when needed. If you encounter this, then it is best to construct your host classes in such a way that they can recover gracefully in these situations.
```
Griffin-Sullivan, line 255:
```suggestion
**Note:** One important thing to keep in mind is that Broker will strip any non-pickleable attributes from Host objects when needed. If you encounter this, then it is best to construct your host classes in such a way that they can recover gracefully in these situations.
```

broker/broker.py
JacobCallahan: ahh i think that was a merge conflict resolution!!

JacobCallahan:
@swadeley @Griffin-Sullivan thanks for the reviews! The latest update should include and/or address everything so far

README.md
```diff
@@ -17,25 +17,32 @@ cp broker_settings.yaml.example broker_settings.yaml
 ```
 Then edit the broker_settings.yaml file
 
+If you are using the Container provider, then install the extra dependency based on your container runtime of choice.
+```
+pip install broker[podman]
+or
+pip install broker[docker]
```
spusateri, line 24: For some reason these commands fail when using `zsh` , I think its worth mentioning you may need to switch to `bash` for them to install correctly

PR #155, opened 2022-06-17
Load env file for dynaconf

mshriver:
There is an issue with dynaconf - still being discussed and written as
an RFE upstream - where consumers of broker are using `.env` files and
configuring their dynaconf instances to read this file.

This causes the broker dynaconf instance to have a partial config based
on the `.env` file from the broker consumer, resulting in a broken
broker dynaconf instance.

In the particular case, robottelo is using broker and is storing
dynaconf vault config within the .env file. The vault config is not
complete, and within broker the settings instance throws exceptions
trying to read from vault.

I would consider this change a workaround for the behavior, not a real
resolution.

It should not interfere with the direct use of broker, or the use of
broker by consumers that are not using vault and not using an env file.

### DIFF
--- a/broker/settings.py
+++ b/broker/settings.py
@@ -25,6 +25,8 @@ settings = Dynaconf(
     settings_file=str(settings_path.absolute()),
     ENVVAR_PREFIX_FOR_DYNACONF="BROKER",
     validators=validators,
+    load_dotenv=True,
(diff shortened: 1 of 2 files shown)


jyejare (approved):

JacobCallahan (approved):
ACK, nice!

PR #157, opened 2022-07-12
Add workaround to disable dynaconf vault loading

JacobCallahan:
This change is needed due to dynaconf #751. This happens when broker is
being used in a project that is using dynaconf's vault integration.

Also reverted the previous try for good measure

### DIFF
--- a/tests/test_settings.py
+++ b/tests/test_settings.py
@@ -9,7 +9,6 @@ def test_default_settings():
     test_provider = TestProvider()
     assert test_provider.instance_name == "default"
     assert test_provider.foo == "bar"
-    assert test_provider.config == "something"
 
 
 def test_alternate_settings():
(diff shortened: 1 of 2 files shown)


tests/test_settings.py
```diff
@@ -24,43 +32,42 @@ def test_validator_trigger():
...
     """
     os.environ["BROKER_TESTPROVIDER__INSTANCES__TEST2__foo"] = "bar"
-    @request.addfinalizer
-    def _clean():
-        del os.environ["BROKER_TESTPROVIDER__INSTANCES__TEST2__foo"]
-
     test_provider = TestProvider(TestProvider="test2")
     assert test_provider.instance_name == "test2"
     assert test_provider.foo == "baz"
+    del os.environ["BROKER_TESTPROVIDER__INSTANCES__TEST2__foo"]
```
mshriver, line 43: While request.addfinalizer might be ugly, it prevents these tests from interfering with each other in the case of an assert failing and pytest never executing this `del` statement.

JacobCallahan: ahhh i see! was only considering the new test

JacobCallahan: Went ahead and just added a parametrized fixture to handle all of that now

tests/test_settings.py
```diff
@@ -1,15 +1,30 @@
 import os
+import sys
 import pytest
 from dynaconf import ValidationError
 from broker.exceptions import ConfigurationError
 from broker.providers.test_provider import TestProvider
 
 
+@pytest.fixture
+def set_envars(request):
```
mshriver, line 10: :heart:

mshriver (approved):

jyejare (approved):

PR #163, opened 2022-08-12
Simulate SSH Session functionality with Container-based Hosts

JacobCallahan:
This change adds the ability to use "sftp" actions against containers.
This is done by (un)packing files in tar archives, which can be added/copied from running containers.
There isn't a current implementation to emulate an interactive shell, but that is likel possible with attach_socket.

I also made some unrelated optimizations to existing code.

### DIFF
--- a/broker/binds/containers.py
+++ b/broker/binds/containers.py
@@ -1,16 +1,17 @@
 class ContainerBind:
-    def __init__(self, host=None, username=None, password=None, port=22):
+    def __init__(self, host=None, username=None, password=None, port=22, timeout=None):
         self.host = host
         self.username = username
         self.password = password
         self.port = port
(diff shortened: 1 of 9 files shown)


broker/binds/containers.py
```diff
@@ -74,24 +75,29 @@ def __repr__(self):
...
+                    **kwargs
+                )
+            )
 
 
 class DockerBind(ContainerBind):
-    def __init__(self, host=None, username=None, password=None, port=2375):
-        super().__init__(host, username, password, port)
+    def __init__(self, port=2375, **kwargs):
+        kwargs["port"] = port
```
mshriver, line 95: always override even if passed?

mshriver (approved):

PR #167, opened 2022-09-14
Improve support for Ansible Tower inventories

JacobCallahan:
This change allows for smarter inventory selection when checking in AnsibleTower-based hosts. Additionally, it improves the restoration of provider instance information. Finally, it simplifies provider settings validation now that Dynaconf supports it.


broker/providers/ansible_tower.py
```diff
@@ -169,6 +177,22 @@ def _set_attributes(self, host_inst, broker_args=None):
...
+    def _get_inventory_by_name(self, inventory_name):
+        if inventory_info := self.v2.inventory.get(search=inventory_name):
+            if inventory_info.count > 1:
+                raise exceptions.ProviderError(
+                    provider="AnsibleTower",
+                    message=f"Ambigious AnsibleTower inventory name {inventory_name}",
+                )
+            elif inventory_info.count == 1:
+                inv_struct = inventory_info.results.pop()
+                return inv_struct.id
```
Griffin-Sullivan, line 189: So is there a case where you can return multiple inventories? How does that happen? Partial matching?

Griffin-Sullivan (approved):

broker/providers/ansible_tower.py
JacobCallahan: Since it performs a search, if the given name is too broad, then we could get multiple results.
For example, specifying an inventory name of `rhv` would likely return 2 inventories

ogajduse (commented):
LGTM after some testing. 

Steps I did:
* `broker checkout --workflow deploy-base-rhel --deploy_rhel_version 8 --notes Requested by broker`
* `cp inventory.yaml inv_before_sync.yaml`
* removed inventory contents ` > inventory.yaml` 
* `broker inventory --sync AnsibleTower`

And looking at the diff, I see that broker did not translate the inventory ID to its name. Can this be changed so we store human-understandable names instead of IDs?

```diff
$ diff -u --color inv_before_sync.yaml inventory.yaml 
--- inv_before_sync.yaml	2022-09-15 21:19:05.099432655 +0200
+++ inventory.yaml	2022-09-15 21:22:18.922426013 +0200
@@ -1,21 +1,88 @@
 - _broker_args:
-    deploy_rhel_version: '8.6'
-    fqdn: ip-10-1-1-1.someopenstack.com
+    _broker_origin: broker_cli:ogajduse
+    deploy_rhel_version: '8'
+    deploy_scenario: baserhel
     host_type: host
-    name:
-    - ogajduse-rhel-8.6-220915T185718762
     notes: Requested by broker
-    os_distribution: RedHat
-    os_distribution_version: '8.6'
-    reported_devices:
-      nics:
-      - eth0
-      - lo
-    template: tpl-rhel-8.6-ga-1663165829
-    tower_inventory: osp-rhos01-satellite-ci
+    tower_inventory: 6
     workflow: deploy-base-rhel
   _broker_provider: AnsibleTower
   _broker_provider_instance: mr
-  hostname: ip-10-1-1-1.someopenstack.com
+  hostname: null
   name: ogajduse-rhel-8.6-220915T185718762
   type: host
```

JacobCallahan:
@ogajduse so that's an interesting question. Should we store the inventories by their name or by their id? I'd like to know what people think about this.
My initial thoughts are:
**Pros**

- It is more readable
- It could solve a potential issue of changing inventory ids

**Cons**

- More difficult on Broker (multiple conversions)
- Could introduce a problem of changing inventory names (with static ids)

ogajduse:
@JacobCallahan 
Can broker store both ID and name on the same level?
```yaml
tower_inventory: osp-rhos01-satellite-ci
tower_inventory_id: 6
```
So broker will use IDs and names would be there for people.

JacobCallahan:
@ogajduse it could, but I likely wouldn't let that happen. I'm leaning more in favor of the name instead of id.

ogajduse (approved):
ACK. Geat improvements! No regressions found (for now :speak_no_evil: ) :+1:

bherrin3 (approved):
While slightly outside of the scope of this MR, it should be noted that hybrid deployments / inventories will now be a thing.

`checkin` works, no matter the default inventory selected for the instance of AnsibleTower

There might be a discussion about making `--tower_inventory` a mandatory or "fully supported" parameter within the `broker` documentation. That way, as we have 6-7 inventories, it is easily obvious how to select different inventories on the CLI.

broker/providers/ansible_tower.py
```diff
@@ -169,6 +176,30 @@ def _set_attributes(self, host_inst, broker_args=None):
             }
         )
 
+    def _translate_inventory(self, inventory):
+        if isinstance(inventory, int):  # already an id, silly
```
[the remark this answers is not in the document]
JacobCallahan: We'd hit an attribute error here if we perform the check as-is and someone put an integer in the inventory. Right now I'm leaning more towards "That would be a user error not worth checking", but if it becomes an issue, we can undoubtedly revisit.

JacobCallahan:
Pending any significant findings, this is likely the last update to this PR. Definitely saw quite a bit of scope creep!

tests/conftest.py
```diff
@@ -0,0 +1,17 @@
+import os
+import pytest
+
+
+@pytest.fixture
+def set_envars(request):
+    """Set and unset one or more envrionment variables"""
```
Griffin-Sullivan, line 7:
```suggestion
    """Set and unset one or more environment variables"""
```
I'm starting to replace @swadeley

Griffin-Sullivan (approved):

tests/conftest.py
Griffin-Sullivan: Also I hope that wasn't Copilot...

ogajduse (approved):
Works for me with the patch that fixed the IndexError. Thanks, @JacobCallahan.
Could you add a test for the `flatten_dict` function?

tests/conftest.py
JacobCallahan: haha nope. all me!

PR #170, opened 2022-10-13
Solve an issue with MP container checkouts

JacobCallahan:
An issue was identified where the _cont_inst property was being stripped from the modified Host class during MP checkouts.
The solution I opted for was to recreate the missing container instance post-pickle. This removed the entire need for the _cont_inst property.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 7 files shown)


Griffin-Sullivan (approved):

adarshdubey-star:
Breaking for me with 
`E           concurrent.futures.process.BrokenProcessPool: A process in the process pool was terminated abruptly while the future was running or pending.`

adarshdubey-star (approved):
Thanks, @JacobCallahan for the prompt fix, LGTM tested locally as well!

PR #171, opened 2022-10-28
Add optimizations for concurrency

JacobCallahan:
Switch ProcessPoolExecutor to ThreadPoolExecutor since these tasks are IO bound. Additionally, experiment with class and session caching.

### DIFF
--- a/broker/providers/ansible_tower.py
+++ b/broker/providers/ansible_tower.py
@@ -3,7 +3,7 @@ import inspect
 import json
 import yaml
 from urllib import parse as url_parser
-from functools import cached_property
+from functools import cache, cached_property
 from dynaconf import Validator
 from broker import exceptions
 from broker.helpers import find_origin, results_filter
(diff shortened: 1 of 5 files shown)


broker/providers/ansible_tower.py
```diff
@@ -22,6 +22,51 @@
 from broker import helpers
 
 
+@cache
+def get_awxkit_and_uname(
```
mshriver, line 26: Nice, cleans things up a bit and gets cached.

JacobCallahan: Yeah, it will help cut back on excessive auths! May need to add something similar for the Container provider as well.

JacobCallahan: Added caching to the container provider's runtime as well. I think we should be good to gut parts of Broker that were added to design around pickle, but I may save that for another time just to play it safe.

JacobCallahan:
Filter changes slipped in between branches, so I'll need to revert those changes before this is ready.

broker/providers/container.py
```diff
@@ -32,6 +33,19 @@ def _host_release():
     caller_host._checked_in = True
 
 
+@cache
+def get_runtime(
```
mshriver, line 37: Why this instead of a `from_dict` classmethod on the class to be instantiated?

mshriver (approved):
approval based on visual inspection, not functional testing.

broker/providers/ansible_tower.py
```diff
@@ -555,6 +567,7 @@ def nick_help(self, **kwargs):
             ]
             if res_filter := kwargs.get("results_filter"):
                 inv = results_filter(inv, res_filter)
+                inv = inv if isinstance(inv, list) else [inv]
```
Griffin-Sullivan, line 570: Do we know what `inv` will be if not a list? If it's a tuple we use `list()`, otherwise I guess wrapping the variable in `[]` is safest.

Griffin-Sullivan (approved):
Ack is also based on code review. Waiting on functional results.

broker/providers/ansible_tower.py
JacobCallahan: yes, it could be a string. this results in treating a single string as a sequence later on.

broker/providers/container.py
JacobCallahan: aside from the additional abstraction involved in that, not much. but there would likely need to be some checking of the dict passed in (specifically keys)

PR #173, opened 2022-11-11
Escape command quotes

peterdragun:
The command should be escaped in case it contains single quotes.
e.g. Registration command contains single quotes, without escaping it is actually not working correctly.


ogajduse (approved):
Nice addition. 
I just wonder if we can use the tempfile and sftp_write approach instead of the `/bin/bash -c` one?

```python
if "'" in command or any([s in command for s in "|&><"]):
    work_with_temp_file_here
```

or better

```python
if any([s in command for s in "|&><'"]):
    work_with_temp_file_here
```

peterdragun:
Yes, that would make sense, I just wanted to create the file just in case when it is really necessary. 

@JacobCallahan What do you think about this?

rplevka (approved):

PR #177, opened 2023-01-13
Add deprecation message for Ansible Tower provider

JacobCallahan:
Our AAP instance is moving away from username/password auth. Users will soon be required to use a token.

### DIFF
--- a/broker/providers/ansible_tower.py
+++ b/broker/providers/ansible_tower.py
@@ -58,7 +58,17 @@ def get_awxkit_and_uname(
             )
     else:  # dynaconf validators should have checked that either token or password was provided
         helpers.emit(auth_type="password")
-        logger.info("Using username and password authentication")
+        if datetime.now() < datetime(2023, 2, 6):
+            time_based_modifier = " and will be unavailable soon"
+        else:
+            time_based_modifier = ""
+        logger.warning(
+            f"Password-based authentication is deprecated{time_based_modifier}. "
+            "Please use a token instead.\n"
+            "See https://docs.ansible.com/automation-controller/latest/html/userguide/"
+            "applications_auth.html#applications-tokens for more information"
+        )
+
         config.credentials = {"default": {"username": uname, "password": pword}}
         config.use_sessions = True
         root.load_session().get()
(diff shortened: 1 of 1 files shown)


Griffin-Sullivan:
Oh boy now I gotta set up token auth to test this works

broker/providers/ansible_tower.py
```diff
@@ -58,7 +58,17 @@ def get_awxkit_and_uname(
...
+        logger.warning(
+            f"Password-based authentication is deprecated{time_based_modifier}. "
+            "Please use a token instead.\n"
+            "See https://docs.ansible.com/automation-controller/latest/html/userguide/"
+            "applications_auth.html#applications-tokens for more information"
+        )
+
         config.credentials = {"default": {"username": uname, "password": pword}}
         config.use_sessions = True
         root.load_session().get()
```
Griffin-Sullivan, line 74: So will you just leave this in and when it's deprecated it won't work? Will it throw an error that we would want to suppress so the user just sees the logger warning?

Griffin-Sullivan (approved):
Results (First command was with username and password, second was username and password commented out and added the token to my settings):
```
broker checkout --nick rhel7                                                                                                                                                                
[INFO 230113 15:31:37] Using provider AnsibleTower to checkout
[WARNING 230113 15:31:37] Password-based authentication is deprecated and will be unavailable soon. Please use a token instead.
    See https://docs.ansible.com/automation-controller/latest/html/userguide/applications_auth.html#applications-tokens for more information
[INFO 230113 15:31:42] Waiting for job: 

broker checkout --nick rhel7                                                                                                                                                   
[INFO 230113 15:34:55] Using provider AnsibleTower to checkout
[INFO 230113 15:34:55] Using token authentication
[INFO 230113 15:35:00] Waiting for job: 
```

broker/providers/ansible_tower.py
JacobCallahan: Haven't decided yet. May leave it in since the functionality would still work against AT/AAP instances that allow password-based auth. For our org, that won't be the case.

rplevka:
an unauthenticated request to `<aap_server>/api/` resource
returns the following info:

```
{
    "description": "AWX REST API",
    "current_version": "[/api/v2/](***/api/v2/)",
    "available_versions": {
        "v2": "[/api/v2/](***/api/v2/)"
    },
    "oauth2": "[/api/o/](***/api/o/)",
    "custom_logo": "",
    "custom_login_info": "",
    "login_redirect_override": ""
}
```

Wouldn't it be better if broker fetched the `custom_login_info` value and display it as warning if set?
This way, we could keep the login-specifics to the individual provider instances instead of hardocding stuff.

not a blocker, just some food for thought

PR #182, opened 2023-01-18
Let sftp_read return data

ogajduse:
This PR allows the user to get the `bytes` object without writing the data to the system.

I have used `sftp_read` method instead of creating a new one.

## TODO
- [x] implement the same in `ContainerSession`
- [x] create the test coverage

To be used in SatelliteQE/robottelo#10547

### DIFF
--- a/broker/session.py
+++ b/broker/session.py
@@ -77,24 +77,27 @@ class Session:
         channel = self.session.open_session()
         return InteractiveShell(channel, pty)
 
-    def sftp_read(self, source, destination=None):
-        """read a remote file into a local destination"""
-        if not destination:
-            destination = source
-        elif destination.endswith("/"):
-            destination = destination + Path(source).name
-        # create the destination path if it doesn't exist
-        destination = Path(destination)
-        destination.parent.mkdir(parents=True, exist_ok=True)
-        destination.touch()
+    def sftp_read(self, source, destination=None, return_data=False):
+        """read a remote file into a local destination or return a bytes object if return_data is True"""
+        if not return_data:
+            if not destination:
+                destination = source
+            elif destination.endswith("/"):
+                destination = destination + Path(source).name
+            # create the destination path if it doesn't exist
+            destination = Path(destination)
+            destination.parent.mkdir(parents=True, exist_ok=True)
         # initiate the sftp session, read data, write it to a local destination
         sftp = self.session.sftp_init()
         with sftp.open(
             source, ssh2_sftp.LIBSSH2_FXF_READ, ssh2_sftp.LIBSSH2_SFTP_S_IRUSR
         ) as remote:
-            with destination.open("wb") as local:
-                for size, data in remote:
-                    local.write(data)
+            captured_data = bytes()
+            for rc, data in remote:
+                captured_data += data
+            if return_data:
+                return captured_data
+            destination.write_bytes(data)
 
     def sftp_write(self, source, destination=None, ensure_dir=True):
         """sftp write a local file to a remote destination"""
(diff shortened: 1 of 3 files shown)


lpramuk (approved):

broker/session.py
```diff
@@ -77,24 +77,30 @@ def shell(self, pty=False):
...
+    def sftp_read(self, source, destination=None, return_data=False):
         """read a remote file into a local destination"""
-        if not destination:
-            destination = source
-        elif destination.endswith("/"):
-            destination = destination + Path(source).name
-        # create the destination path if it doesn't exist
-        destination = Path(destination)
-        destination.parent.mkdir(parents=True, exist_ok=True)
-        destination.touch()
```
ogajduse, line 89: I have removed the touch. I see no point in keeping it there as the file would get created anyway when we open and write to it.

ogajduse:
@JacobCallahan How about the test coverage for `sftp_read`? Do you want me to create it as a part of this PR?

JacobCallahan:
> @JacobCallahan How about the test coverage for `sftp_read`? Do you want me to create it as a part of this PR?

Yes, you can add some tests to the functional directory. Those operate on live hosts.

rplevka (approved):

broker/session.py
```diff
@@ -77,24 +77,30 @@ def shell(self, pty=False):
...
-                    local.write(data)
+            if return_data:
+                captured_data = bytes()
+                for rc, data in remote:
+                    captured_data += data
+                return captured_data
+            else:
+                with destination.open("wb") as local:
+                    for rc, data in remote:
+                        local.write(data)
```
JacobCallahan, line 103: We likely can simplify this a bit more. Since we are returning as bytes, we can read it into the object then either return that object or write it to the destination file.
```suggestion
            captured_data = bytes()
            for rc, data in remote:
                captured_data += data
            if return_data:
                return captured_data
            destination.write_bytes(data)
```

JacobCallahan (changes requested):

broker/session.py
```diff
@@ -77,24 +77,30 @@ def shell(self, pty=False):
         channel = self.session.open_session()
         return InteractiveShell(channel, pty)
 
-    def sftp_read(self, source, destination=None):
+    def sftp_read(self, source, destination=None, return_data=False):
```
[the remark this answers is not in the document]
ogajduse: Done

broker/session.py
```diff
@@ -257,20 +260,30 @@ def sftp_write(self, source, destination=None, ensure_dir=True):
...
         data, status = self._cont_inst._cont_inst.get_archive(source)
         logger.debug(f"{self._cont_inst.hostname}: {status}")
         all_data = b"".join(d for d in data)
         if destination.name == "_raw":
             return all_data
         with helpers.data_to_tempfile(all_data, as_tar=True) as tar:
-            logger.debug(f"Extracting {source} to {destination}")
-            tar.extractall(destination.parent if destination.is_file() else destination)
+            if len(tar.getmembers()) == 1:
+                f = tar.extractfile(tar.getmember(Path(source).name))
```
ogajduse, line 274: Should we be worried about triplication of the data in the memory in case we transfer huge files?
See `data`, `all_data`, `f`.

JacobCallahan: could overwrite `data` on the join line. would at least help a bit. may also be able to `del all_data` in the line after we enter the context manager, but I don't know how it uses that reference

tests/conftest.py
```diff
@@ -15,3 +17,30 @@ def set_envars(request):
...
+        ), "Local file is different from the received one"
+        assert (
+            loc_settings_path.read_bytes() == data
+        ), "Local file is different from the received one (return_data=True)"
+        assert data == Path(tmp.file.name).read_bytes(), "Received files do not match"
+
+
+@pytest.fixture(params=[functional_host_actions])
+def host_actions(request):
+    return request.param
```
JacobCallahan, line 46: ehhh setup is one thing, but I personally don't like moving test logic outside of the tests themselves. The tests can also differ in the breadth of functionality as well, which might be the better course of action. For example one test (per Provider) that tests all of the file actions.

JacobCallahan (changes requested):

broker/session.py
ogajduse: Say I re-use `data` to store `all_data` as you suggested.
The reference to `data` is not used once we enter the context manager. So I tried to delete it as you suggested.
Using that patch I have successfully tested `ContainerSession.sftp_read` method using the `test_container_e2e_mp` test.
I will push a commit with this patch soon.

tests/conftest.py
ogajduse: @JacobCallahan 
I agree, I can definitely improve this one.
I can imagine having one test per Provider only in the case of the Container provider. Hosts created using the AnsibleTower provider add significant time to the test time. The time it takes to spawn a host utilizing the Container provider is negligible compared to AnsibleTower.

For that reason, I would prefer having one functional test per provider as we do not "waste" time deploying instances.

Q1: Would you prefer code duplication over my solution? :smile: 

Q2: Aren't all Host objects supposed to behave roughly the same no matter the provider they are deployed with?

broker/session.py
ogajduse: Modified in a7acaa4

tests/functional/test_containers.py
```diff
@@ -83,7 +97,20 @@ def test_container_e2e_mp():
...
+                data = c_host.session.sftp_read(
+                    f"{remote_dir}/{loc_settings_path.name}", return_data=True
+                )
+                assert (
+                    loc_settings_path.read_bytes() == Path(tmp.file.name).read_bytes()
+                ), "Local file is different from the received one"
+                assert (
+                    loc_settings_path.read_bytes() == data
+                ), "Local file is different from the received one (return_data=True)"
+                assert data == Path(tmp.file.name).read_bytes(), "Received files do not match"
```
JacobCallahan, line 116: looks like you're just missing a newline
```suggestion
                assert data == Path(tmp.file.name).read_bytes(), "Received files do not match"

```

JacobCallahan (changes requested):

tests/functional/test_containers.py
ogajduse: good catch!

ogajduse:
```
$ pytest tests/functional/test_containers.py -k test_container_e2e
========================================================================= test session starts =========================================================================
platform linux -- Python 3.11.1, pytest-7.2.1, pluggy-1.0.0 -- /home/ogajduse/.virtualenvs/broker-dev/bin/python
cachedir: .pytest_cache
rootdir: /home/ogajduse/repos/broker, configfile: pytest.ini
collected 8 items / 6 deselected / 2 selected                                                                                                                         

tests/functional/test_containers.py::test_container_e2e PASSED                                                                                                  [ 50%]
tests/functional/test_containers.py::test_container_e2e_mp PASSED                                                                                               [100%]

============================================================= 2 passed, 6 deselected in 85.14s (0:01:25) ==============================================================
```

JacobCallahan (approved):
ACK, nice!

PR #185, opened 2023-01-26
[0.3.0] Refactor Provider registration system

JacobCallahan:
This new version is more in-line with a plugin-based approach.
A Provider author can now create a new Provider class and register it
without having to touch any other files but their own.

To do this, we now have automatic registration of any class that inherits
from the base Provider class. Additionally, registering actions is done via
a new decorator.

Also

Remove PickleSafe base class
Broker has moved from multiprocessing to multithreading for concurrency.
Because of this, we no longer have to worry about pickling objects.

### DIFF
--- a/broker/broker.py
+++ b/broker/broker.py
@@ -1,28 +1,12 @@
 from logzero import logger
-from broker.providers.ansible_tower import AnsibleTower
-from broker.providers.container import Container
-from broker.providers.test_provider import TestProvider
+from broker.providers import PROVIDERS, PROVIDER_ACTIONS, _provider_imports
 from broker.hosts import Host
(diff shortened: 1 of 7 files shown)


broker/hosts.py
```diff
@@ -1,12 +1,11 @@
 # from functools import cached_property
 from logzero import logger
 from broker.exceptions import NotImplementedError
-from broker.helpers import PickleSafe
 from broker.session import ContainerSession, Session
 from broker.settings import settings
 
 
-class Host(PickleSafe):
+class Host():
```
JacobCallahan, line 8:
```suggestion
class Host:
```

PR #188, opened 2023-01-27
[0.3.0] Complete rewrite of filtering

JacobCallahan:
This commit completely rewrites the filtering system. It is now based on looser python expressions against list objects. This allows for more flexibility in the filtering system. It also allows for more complex filtering expressions.

I've also used some of the MockStub object's magic to allow for attribute access on the more complex filter subjects, like inventory hosts.


broker/helpers.py
```diff
@@ -91,58 +91,34 @@ def flatten_dict(nested_dict, parent_key="", separator="_"):
...
+                    ),
+                    filter_list,
+                )
+            )
+    return [dict(item) if isinstance(item, MockStub) else item for item in filter_list]
+
+
+def get_inventory_attr(filtered_inventory, attr):
+    """Get an attribute from the inventory"""
+    return [eval(attr[1:], {"host": host}) for host in filtered_inventory]
```
Griffin-Sullivan, line 121: I don't see this being used anywhere. Is this a "just in case we need it" helper?

tests/test_helpers.py
```diff
@@ -110,3 +115,36 @@ def test_flatten_duplicate():
...
+    """Test that a python list copy operation returns all entries in the inventory"""
+    filtered = helpers.eval_filter(fake_inventory, "@inv[:]")
+    assert len(filtered) == 10
+
+
+def test_eval_filter_list_last(fake_inventory):
+    """Test that a neative list index returns the last host in the inventory"""
+    filtered = helpers.eval_filter(fake_inventory, "@inv[-1]")
+    assert len(filtered) == 1
+    assert filtered[0]["hostname"] == "dhcp-369.test.example.com"
```
Griffin-Sullivan, line 130: Is there any way we could load the fake inventory file into a list of constants for this module to use? If that inventory file changes we will need to update these hard coded hostnames.

Griffin-Sullivan (approved):
```
======================= test session starts=========================
platform linux -- Python 3.11.1, pytest-7.2.1, pluggy-1.0.0 -- /home/gsulliva/Programming/venvbroker/bin/python3
cachedir: .pytest_cache
configfile: pytest.ini
collected 17 items / 12 deselected / 5 selected                                                                                                                                       

tests/test_helpers.py::test_eval_filter_list_copy PASSED                                                                                                                        [ 20%]
tests/test_helpers.py::test_eval_filter_list_last PASSED                                                                                                                        [ 40%]
tests/test_helpers.py::test_eval_filter_list_slice PASSED                                                                                                                       [ 60%]
tests/test_helpers.py::test_eval_filter_attribute PASSED                                                                                                                        [ 80%]
tests/test_helpers.py::test_eval_filter_chain PASSED                                                                                                                            [100%]
================= 5 passed, 12 deselected in 0.34s ================

tests/test_helpers.py
JacobCallahan: that's an option, but selecting the proper attributes may be a bit messy. I'm not sure that it would be worth it.

broker/helpers.py
JacobCallahan: good question. I think this may be before I loaded the inventory hosts as mockstub objects.

JacobCallahan: removed!

PR #189, opened 2023-02-08
Fix logging by using a custom filter only for the broker logger

ogajduse:


broker/logger.py
```diff
@@ -20,8 +20,20 @@ class LOG_LEVEL(IntEnum):
     ERROR = logging.ERROR
 
 
+class RedactingFilter(logging.Filter):
```
jyejare, line 23: Please add description as to what it does !

broker/logger.py
```diff
@@ -20,8 +20,20 @@ class LOG_LEVEL(IntEnum):
...
+    def __init__(self, sensitive):
+        super(RedactingFilter, self).__init__()
+        self._sensitive = sensitive
+
+    def filter(self, record):
+        if isinstance(record.args, dict):
+            record.args = redact_dynaconf(record.args)
+        else:
```
jyejare, line 31: Good to check the instance type here as well for tuple or any other Mapping type.

broker/logger.py
```diff
@@ -137,11 +136,7 @@ def setup_logzero(
     patch_awx_for_verbosity(awxkit.api)
     set_log_level(level)
     set_file_logging(file_level, path)
-    global _old_factory
-    lrf = logging.getLogRecordFactory()
-    if lrf.__name__ is not broker_record_factory.__name__:
-        _old_factory = lrf
-        logging.setLogRecordFactory(broker_record_factory)
+    logzero.logger.addFilter(RedactingFilter(_sensitive))
```
jyejare, line 139: This `readact_dynaconf` is calling `RedactingFilter` and `RedactingFilter` is calling this function. This will end up with a recursive loop.

jyejare, line 139: This `readact_dynaconf` is calling `RedactingFilter` and `RedactingFilter` is calling this function. I believe this will end up with a recursive loop.

jyejare (changes requested):

broker/logger.py
ogajduse: It could be either tuple or dict.
`args (tuple | dict[str, Any]) – Variable data to merge into the msg argument to obtain the event description.`
See https://docs.python.org/3/library/logging.html#logging.LogRecord

broker/logger.py
ogajduse: `This readact_dynaconf is calling RedactingFilter`  not true. `redact_dynaconf` calls itself recursively, yes, but it does not call `RedactingFilter`. How could you even call a class? :thinking: 

But thanks for reminding me that I wanted to move `redact_dynaconf` to the filter class which I will do in the next commit.

ogajduse:
In robottelo using broker from the PR branch:
```
(Pdb++) from robottelo.logging import logger
(Pdb++) logger.filters
[]
```
This means that the filter that is used for broker logger is not used for the robottelo logger.

broker/logger.py
jyejare: ahh, looks like my github view of this PR shown somwthing wrong

Griffin-Sullivan:
I can confirm I ran a test (test_rhcloud_inventory_without_manifest) with broker 0.2.12 and it failed with the typeerror, and then installing this branch fixed the issue.

Griffin-Sullivan (approved):
I'm not gonna act like I'm 100% on the logic here, but I can at least confirm that this fixed the TypeError issue in robottelo.

broker/logger.py
```diff
@@ -20,8 +20,39 @@ class LOG_LEVEL(IntEnum):
...
+        """
+        if isinstance(data, (list, tuple)):
+            data_copy = [self.redact_dynaconf(item) for item in data]
+        elif isinstance(data, dict):
+            data_copy = copy.deepcopy(data)
+            for k, v in data_copy.items():
+                if isinstance(v, (dict, list)):
+                    data_copy[k] = self.redact_dynaconf(v)
+                elif k in self._sensitive and v:
+                    data_copy[k] = "*" * 6
```
JacobCallahan, line 49: Much easier to just write out the string
```suggestion
                    data_copy[k] = "******"
```

JacobCallahan (changes requested):
One small request

JacobCallahan (approved):
ACK, nice

PR #193, opened 2023-03-02
Allow for additional arguments to be passed to provider help commands

JacobCallahan:
This change allows users to pass along additional arguments to provider help commands. Providers can then choose to use these arguments as they see fit.

One new possible example is to specify a tower_inventory when listing the available templates for AnsibleTower.

I also added a couple of new logging statements to make it clear what tower inventory is being used by Broker and when.

### DIFF
--- a/broker/providers/ansible_tower.py
+++ b/broker/providers/ansible_tower.py
@@ -348,10 +348,15 @@ class AnsibleTower(Provider):
             return None
 
     def _compile_host_info(self, host):
+        # attempt to get the hostname from the host variables and then facts
+        if not (hostname := host.variables.get("fqdn")):
+            if hasattr(host_facts := host.related.ansible_facts.get(), "results"):
+                hostname = host_facts.results[0].ansible_facts.get("ansible_fqdn")
         host_info = {
             "name": host.name,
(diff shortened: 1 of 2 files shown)


broker/providers/ansible_tower.py
```diff
@@ -351,7 +351,7 @@ def _compile_host_info(self, host):
         host_info = {
             "name": host.name,
             "type": host.type,
-            "hostname": host.variables.get("fqdn"),
+            "hostname": host.variables.get("fqdn", host.facts.get("ansible_fqdn")),
```
ColeHiggins2, line 354: this will be very helpful!

ColeHiggins2 (approved):
Looks good to me !

broker/providers/ansible_tower.py
JacobCallahan: i hope so

ColeHiggins2: after further investigation. It looks like 'Host' object has no attribute 'facts'.

Griffin-Sullivan (approved):

ogajduse (approved):

PR #194, opened 2023-03-02
[0.3.0] Make Broker more opinionated about where its files are located

### LINKED ISSUE #190: Make inventory_file have a sane default, and not required
While trying to put together a minimal settings file, I noticed that an error is thrown if I do not set `inventory_file: inventory.yaml` in `broker_settings.yaml`. This feels like a sane default that could be introduced thus allowing it to be removed from a minimal configuration.

JacobCallahan:
This change creates a new standard for where Broker files are located. By default, Broker will be based in ~/.broker/ but will still listen for the BROKER_DIRECTORY environment variable.

As a result, Broker will no longer honor any alternative locations for its inventory file.

Fixes #190

### DIFF
--- a/broker/settings.py
+++ b/broker/settings.py
@@ -4,20 +4,22 @@ from dynaconf import Dynaconf, Validator
 from dynaconf.validator import ValidationError
 from broker.exceptions import ConfigurationError
 
-settings_file = "broker_settings.yaml"
-BROKER_DIRECTORY = Path()
+BROKER_DIRECTORY = Path.home().joinpath(".broker")
 
 if "BROKER_DIRECTORY" in os.environ:
     envar_location = Path(os.environ["BROKER_DIRECTORY"])
     if envar_location.is_dir():
         BROKER_DIRECTORY = envar_location
 
+# ensure the broker directory exists
+BROKER_DIRECTORY.mkdir(parents=True, exist_ok=True)
+
 settings_path = BROKER_DIRECTORY.joinpath("broker_settings.yaml")
 inventory_path = BROKER_DIRECTORY.joinpath("inventory.yaml")
 
 validators = [
     Validator("HOST_USERNAME", default="root"),
-    Validator("HOST_PASSWORD", must_exist=True),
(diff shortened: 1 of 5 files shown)


broker/settings.py
```diff
@@ -4,20 +4,22 @@
...
 if "BROKER_DIRECTORY" in os.environ:
     envar_location = Path(os.environ["BROKER_DIRECTORY"])
     if envar_location.is_dir():
         BROKER_DIRECTORY = envar_location
 
+# ensure the broker directory exists
+BROKER_DIRECTORY.mkdir(parents=True, exist_ok=True)
+
 settings_path = BROKER_DIRECTORY.joinpath("broker_settings.yaml")
 inventory_path = BROKER_DIRECTORY.joinpath("inventory.yaml")
```
Griffin-Sullivan, line 18: Should these be showing up under my `.broker` directory?

Griffin-Sullivan (approved):
Everything is working fine and exporting the BROKER_DIRECTORY allows me to run broker anywhere. I see the `.broker` directory being created, but I only see `logs/` in there.

ogajduse (approved):
```
$ broker --version
Version: 0.0.0
Broker Directory: /home/ogajduse/.broker
Settings File: /home/ogajduse/.broker/broker_settings.yaml
Inventory File: /home/ogajduse/.broker/inventory.yaml
Log File: /home/ogajduse/.broker/logs/broker.log

$ tree /home/ogajduse/.broker/
/home/ogajduse/.broker/
└── logs
    └── broker.log

2 directories, 1 file

$ broker inventory
[INFO 230303 19:12:07] Pulling local inventory

$ tree /home/ogajduse/.broker/
/home/ogajduse/.broker/
└── logs
    └── broker.log

2 directories, 1 file
```

I do not have BROKER_DIRECTORY` env var set.
```
$ echo $BROKER_DIRECTORY
```

broker/settings.py
ogajduse: I would say that the user is expected to put these two files there.

PR #204, opened 2023-04-03
0.3.0 Release!

JacobCallahan:
This PR brings the 0.3.0 branch into master. The PRs currently included are:

- #185 
- #188 
- #194 
- #203 
- #206 

For those wishing to test these changes, running all functional tests locally is a great start.
From there, mess around with the new filtering system and see if it is performing as expected.
You can also test out the changes to the settings file and Broker's new fallback directory.


JacobCallahan:
Here's the new guide for filtering

**Filter hosts for Broker actions**

Actions that Broker can take against hosts (checkin, duplicate, extend) can take in a filter argument. This filter will decide which hosts the actions are applied to. A filter by itself will not select hosts for these actions, you will still need to specify which hosts to act against, or use `--all` when available. From there, the filter decides which of those hosts make it through to be acted upon.

Broker's filters are based on what is stored in its local inventory file. Therefore, only properties in that file are filter-able. Nested properties are annotated with a `.` notation. For example, a top-level property `hostname` can be accessed by itself. However, a nested property of `_broker_args` called `version` would be accessed by `_broker_args.version`.

Filters are any valid python expression that acts upon a filterable object. A filterable object is a list typically noted with either `@inv` or `@res`. Broker will replace that notation, in the filter expression, with a valid list whose contents are context-dependent.

**Example filters:**

`--filter '"test" in @inv.hosname'` The string test should exist somewhere in the hostname value
`--filter '@inv[-1]'` Use the last host in the inventory
`--filter '@inv[3:7]'` Pick the 3rd through 6th hosts in the inventory
`--filter '@inv._broker_args.template.startswith("deploy-sat")'` The template should start with the string "deploy-sat"

You can also chain multiple filters together, feeding the results of the previous filter into the next, using the pipe `|` symbol. This allows you to perform multiple layers of filtering in a single expression.

`--filter '"test" in @inv.name | @inv._broker_args.provider != "RHEV"'` The host's name should have test in it and the provider should not equal RHEV.

PR #206, opened 2023-04-04
[0.3.0] Add the ability for Broker to download a missing settings file

JacobCallahan:
When the settings file is missing from the user's Broker directory,
Broker will prompt them to download the file from GitHub and open an editor.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -5,6 +5,12 @@ on:
     branches: [master]
   pull_request:
     types: [opened, synchronize, reopened]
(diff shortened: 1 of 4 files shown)


JacobCallahan:
to test this, simply delete (or move) your current broker settings file, then try to run any broker command.

Griffin-Sullivan (approved):
Code looks good and during my testing I was able to move my broker_settings file out of the root directory and when I ran a command it prompted me to download the settings file and remake a new one in ~/.broker. After deleting that new file and moving my original to ~/.broker I was able to sync the inventory and everything is working like normal.

PR #208, opened 2023-04-18
Add contextual awareness to host creation

JacobCallahan:
This change fixes an issue we're seeing where host entries, missing a hostname and ip, are being reconstructed for checkin. While that shouldn't be an issue during checkin, it would be if used for other purposes.

With that in mind, this adds a check to see if the host is being reconstructed. If so, then it will just debug log instead of raising an error.

Add in some new workarounds and log statements
We're seeing some strange behavior in CI. These are attempts to solve them
and/or get more information about what's going on.

### DIFF
--- a/broker/hosts.py
+++ b/broker/hosts.py
@@ -16,7 +16,12 @@ class Host:
         else:
             self.hostname = hostname or kwargs.get("ip", None)
             if not self.hostname:
-                raise HostError("Host must be constructed with a hostname or ip")
+                # check to see if we're being reconstructued, likely for checkin
+                import inspect
+                if any(f.function == "reconstruct_host" for f in inspect.stack()):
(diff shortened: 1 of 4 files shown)


Griffin-Sullivan:
It seems like the checkin is happening just as it normally does. Is that what we want? I'll try again on current latest to see if I get a failure.

Griffin-Sullivan (approved):
From what I gathered, before checking out this PR I ran a checkout, broker sync, change hostname and ip to null, and then check in. That returns the following error: `[ERROR 230501 14:28:38] BrokerError: 'str' object has no attribute 'hostname'`. Based off the description I don't know if this is exactly the issue we are trying to fix here, but after running the same steps with this PR installed, I do not get the error.

PR #211, opened 2023-05-08
Add a tail_file context manager to the Host class

JacobCallahan:
This is a pretty common request and should be useful in a number of scenarios. Usage is pretty simple:
```python
    with my_host.session.tail_file("/var/log/messages") as res:
        # do something that creates new messages
    print(res.stdout)
```

### DIFF
--- a/broker/hosts.py
+++ b/broker/hosts.py
@@ -1,4 +1,3 @@
-# from functools import cached_property
 from logzero import logger
 from broker.exceptions import NotImplementedError, HostError
 from broker.session import ContainerSession, Session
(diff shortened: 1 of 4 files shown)


JacobCallahan:
I was debating whether this should be on the host object itself or on the host's session object. Let me know what you think. The alternative would look like this

```python
    with my_host.session.tail_file("/var/log/messages") as res:
        # do something that creates new messages
    print(res.stdout)
```

JacobCallahan:
@rplevka the new version is in now. The downside I forgot about is that it needed to be duplicated for the ContainerSession, but that's not a big deal.

JacobCallahan:
@Griffin-Sullivan great idea! Is that something you're wanting to do or would you prefer I handle that in this PR?

Griffin-Sullivan:
I can do it in a separate PR.

PR #214, opened 2023-05-15
Add initial support for a Beaker provider.

JacobCallahan:
This change adds a very basic Beaker provider and its bind. For now, this relies on the beaker-client cli and config through that. Currently the only supported checkout/execute method is via job xml.


JacobCallahan:
@bherrin3 this is a VERY EARLY draft and does not work yet, but you can get an idea of where i'm going with it. Many of the pieces are in place, but not enough.

PR #215, opened 2023-05-22
Add Broker.multi_manager

JacobCallahan:
This change introduces a way to carry out multiple kinds of checkouts at the same time when using Broker as a library.

from the docstring:
```
        Given a mapping of names to Broker argument dictionaries:
        create multiple Broker instances, check them out in parallel, yield, then checkin.
        Example:
            with Broker.multi_mode(
                rhel7={
                    "host_class": ContentHost,
                    "workflow": "deploy_base_rhel",
                    "deploy_rhel_version": "7",
                },
                rhel8={
                    "host_class": ContentHost,
                    "workflow": "deploy_base_rhel",
                    "deploy_rhel_version": "8",
                }
            ) as host_dict:
                pass
        All are checked out at the same time. The user is presented with the hosts in
        a dictionary by argument name e.g. host_dict["rhel7"] is a ContentHost object

```

### DIFF
--- a/broker/broker.py
+++ b/broker/broker.py
@@ -1,3 +1,4 @@
+from contextlib import contextmanager
 from logzero import logger
 from broker.providers import PROVIDERS, PROVIDER_ACTIONS, _provider_imports
 from broker.hosts import Host
@@ -22,6 +23,7 @@ class mp_decorator:
 
     The decorated method is expected to return an itearable.
     """
+
     MAX_WORKERS = None
     """ If set to integer, the count of workers will be limited to that amount.
      If set to None, the max workers count of the EXECUTOR will match the count of items."""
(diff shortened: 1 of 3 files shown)


JacobCallahan:
New functional test passing
Screenshot from 2023-05-22 15-30-06

JacobCallahan:
Now, there are some decisions that I'm not final on. Let me know what you all think about these?

1. When there is a single host, as a value per key, separate it from the list.
2. Perform all setup and teardown methods in a thread pool.

JacobCallahan:
Ok, latest update also executes all host setup and teardown tasks in a thread pool

ogajduse (approved):
I like this feature. LGTM.
Great job, @JacobCallahan!

Issue #212, opened 2023-05-10
Test suite needs to be updated to look for broker_settings.yaml in ~/.broker


Griffin-Sullivan:
Currently, the tests are hard coded to search for `broker_settings.yaml` in the root directory of broker. After https://github.com/SatelliteQE/broker/pull/206, we now need to look for the file in `~/.broker/`. See https://github.com/SatelliteQE/broker/pull/211 for an example.

Gauravtalreja1:
@tpapaioa @JacobCallahan Can this issue be closed as completed?

JacobCallahan:
@Gauravtalreja1 on second look, yes. There is a larger issue around config files used in testing that will be resolved later.

ogajduse:
@JacobCallahan Can you please link the PR that resolved it for better traceability?

PR #216, opened 2023-05-26
Updating tests to use real and fake broker_settings.yaml

Griffin-Sullivan:
Fixes https://github.com/SatelliteQE/broker/issues/212. All of the tests that are non-functional will make backup files for your settings and inventory and copy in test settings. Once the test is finished they are reset to normal.

image

### DIFF
--- a/broker/providers/__init__.py
+++ b/broker/providers/__init__.py
@@ -76,6 +76,7 @@ class Provider(metaclass=ProviderMeta):
     _checkout_options = []
     _execute_options = []
     _fresh_settings = settings.dynaconf_clone()
+    _settings_id = settings._broker_settings_id
     _sensitive_attrs = []
 
     def __init__(self, **kwargs):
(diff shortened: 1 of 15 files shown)


tests/conftest.py
```diff
@@ -15,3 +17,21 @@ def set_envars(request):
         os.environ[request.param[0]] = request.param[1]
         yield
         del os.environ[request.param[0]]
+
+@pytest.fixture(scope="module")
+def temp_settings_and_inventory():
```
ogajduse, line 22: What about using the [tmp_path](https://docs.pytest.org/en/7.3.x/reference/reference.html#std-fixture-tmp_path) fixture and creating the new files in it? Then setting `BROKER_DIRECTORY` to `tmp_path` and unsettling it once the test is done?

JacobCallahan:
@Griffin-Sullivan will you have a chance to revisit this sometime soon?

Griffin-Sullivan:
Yeah I'll try to do it later today. I just need to get those 2 tests fixed and add @ogajduse's suggestions.

broker/providers/__init__.py
```diff
@@ -47,13 +47,18 @@ class Provider(metaclass=ProviderMeta):
...
+    _settings_id = settings._id
     _sensitive_attrs = []
 
     def __init__(self, **kwargs):
         self._construct_params = []
         cls_name = self.__class__.__name__
         logger.debug(f"{cls_name} provider instantiated with {kwargs=}")
         self.instance = kwargs.pop(f"{cls_name}", None)
+        if not self._settings_id == settings._id:
+            logger.debug("Settings object has been reloaded, reloading provider settings")
```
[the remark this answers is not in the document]
Griffin-Sullivan: For running the test suite we need to change the settings file being used. Now that we automatically look for settings in `~/.broker` we need to change it to our broker_settings for testing when the tests are running.

tests/conftest.py
Griffin-Sullivan: We could try it if you think that's an improvement over this implementation. @JacobCallahan any thoughts since you helped me with our current solution?

Griffin-Sullivan:
Screw this

PR #221, opened 2023-06-30
Allow remote_copy to copy to a different path

### LINKED ISSUE #165: [RFE] Add a function to transfer directory/files from one host to another at specific location
I know that there is a function `remote_copy` in Broker to move file between hosts but this function only allows user to copy one file to the exact same location to another host.
My suggestion is to make a function that allows users to transfer files or directory recursively from the source host to the target host to the specific location that would be user-defined.
As the function with these features does not exist yet I am forced to use something like this in my tests.
i. e.:
`source_host.execute('sshpass -p "password" scp -o StrictHostKeyChecking=no -r /var/backup/satellite-backup target.hostname.com:/backup/')`

JacobCallahan:
This change allows a user to supply a different destination path for remote_copy.

Fixes #165

### DIFF
--- a/broker/logger.py
+++ b/broker/logger.py
@@ -138,7 +138,8 @@ def setup_logzero(
     path="logs/broker.log",
 ):
     urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
-    patch_awx_for_verbosity(awxkit.api)
+    if isinstance(level, str) and level.lower() == "trace":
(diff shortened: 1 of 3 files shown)


JacobCallahan:
@LadislavVasina1 I believe this should do what you want, but please let me know if you were thinking of something different.

LadislavVasina1:
@JacobCallahan This is exactly what I meant. Looks good to me.

PR #222, opened 2023-07-07
Add Ruff as a code checker

JacobCallahan:
Ruff is a newer code checker that is written in Rust and is very fast. Adding this now with a number of pretty sane checks to start with. I recommend having Ruff integrated into your editor to pick out issues as you write code.

Most of the changes in this PR are adding/modifying docstrings.


JacobCallahan:
Available rules can be found here: https://beta.ruff.rs/docs/rules/

If you find one you think should be added, comment with your reasoning.

JacobCallahan:
All tests passed locally, including functional for all providers.
Screenshot from 2023-07-07 20-21-56

ruff.toml
```diff
@@ -0,0 +1,110 @@
...
+ignore = [
+    "ANN", # flake8-annotations
+    "PGH001", # No builtin eval() allowed
+    "S102", # Use of exec detected
+    "Q000", # Single quotes found but double quotes preferred
+    "D202", # No blank lines allowed after function docstring
+    "D203", # 1 blank line required before class docstring
+    "D213", # Multi-line docstring summary should start at the second line
+    "D406", # Section name should end with a newline
+    "D407", # Section name underlining
```
ogajduse, line 78: Why do we need to ignore D407?

ruff.toml
```diff
@@ -0,0 +1,110 @@
...
+    "E501", # line too long
+    "E731", # do not assign a lambda expression, use a def
+    "PLC1901", # Lots of false positives
+    # False positives https://github.com/astral-sh/ruff/issues/5386
+    "PLC0208", # Use a sequence type instead of a `set` when iterating over values
+    # Ignored due to performance: https://github.com/charliermarsh/ruff/issues/2923
+    "PLR0913", # Too many arguments to function call ({c_args} > {max_args})
+    "RUF012", # Mutable class attributes should be annotated with typing.ClassVar
+    "UP038", # Use `X | Y` in `isinstance` call instead of `(X, Y)`
+    "D107", # Missing docstring in __init__
```
ogajduse, line 88: Why can't `__init__` have a docstring too?

ruff.toml
```diff
@@ -0,0 +1,110 @@
...
+
+ignore = [
+    "ANN", # flake8-annotations
+    "PGH001", # No builtin eval() allowed
+    "S102", # Use of exec detected
+    "Q000", # Single quotes found but double quotes preferred
+    "D202", # No blank lines allowed after function docstring
+    "D203", # 1 blank line required before class docstring
+    "D213", # Multi-line docstring summary should start at the second line
+    "D406", # Section name should end with a newline
```
ogajduse, line 77: What is the reason for ignoring D406?

ruff.toml
```diff
@@ -0,0 +1,110 @@
...
+    "D407", # Section name underlining
+    "E501", # line too long
+    "E731", # do not assign a lambda expression, use a def
+    "PLC1901", # Lots of false positives
+    # False positives https://github.com/astral-sh/ruff/issues/5386
+    "PLC0208", # Use a sequence type instead of a `set` when iterating over values
+    # Ignored due to performance: https://github.com/charliermarsh/ruff/issues/2923
+    "PLR0913", # Too many arguments to function call ({c_args} > {max_args})
```
ogajduse, line 85: Isn't it better to define a sane number of max arguments?
https://beta.ruff.rs/docs/settings/#pylint-max-args

ruff.toml
```diff
@@ -0,0 +1,110 @@
...
+    "D202", # No blank lines allowed after function docstring
+    "D203", # 1 blank line required before class docstring
+    "D213", # Multi-line docstring summary should start at the second line
+    "D406", # Section name should end with a newline
+    "D407", # Section name underlining
+    "E501", # line too long
+    "E731", # do not assign a lambda expression, use a def
+    "PLC1901", # Lots of false positives
+    # False positives https://github.com/astral-sh/ruff/issues/5386
+    "PLC0208", # Use a sequence type instead of a `set` when iterating over values
```
ogajduse, line 83: ruff `0.0.277` still reports this issue even though https://github.com/astral-sh/ruff/releases/tag/v0.0.276 fixed the false positive.

I would like to see this one selected, rather than ignored.

ruff.toml
```diff
@@ -0,0 +1,110 @@
...
+]
+
+ignore = [
+    "ANN", # flake8-annotations
+    "PGH001", # No builtin eval() allowed
+    "S102", # Use of exec detected
+    "Q000", # Single quotes found but double quotes preferred
+    "D202", # No blank lines allowed after function docstring
+    "D203", # 1 blank line required before class docstring
+    "D213", # Multi-line docstring summary should start at the second line
```
ogajduse, line 76: Why can't broker be compliant with pydocstyle recommendations?
I would like to see at least D202 not being ignored.

ruff.toml
```diff
@@ -0,0 +1,110 @@
...
+    "broker",
+]
+combine-as-imports = true
+
+[per-file-ignores]
+
+# Allow for main entry & scripts to write to stdout
+# "homeassistant/__main__.py" = ["T201"]
+# "homeassistant/scripts/*" = ["T201"]
+# "script/*" = ["T20"]
```
ogajduse, line 107: HA is definitely a great source to get inspiration from :heart:

ogajduse (changes requested):

ruff.toml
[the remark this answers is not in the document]
JacobCallahan: i've considered that, but am a bit on the fence about it. right now we don't have many tool that would typically go in a pyproject.toml, this one being a large exception. perhaps if we added pre-commit with black and ruff...

broker/broker.py
```diff
@@ -1,25 +1,47 @@
...
 
 
 def _try_teardown(host_obj):
-    """Try a host's teardown method and return an exception message if needed"""
+    """Try a host's teardown method and return an exception message if needed."""
     try:
         host_obj.teardown()
-    except Exception as err:
+    except Exception as err:  # noqa: BLE001
+        logger.debug(f"Tell Jake the exception was: {err}")
```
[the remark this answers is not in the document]
JacobCallahan: ha yep! gotta see what these are in automation

broker/helpers.py
```diff
@@ -123,24 +124,25 @@ def dict_from_paths(source_dict, paths):
...
-            filter_list = filter_list if isinstance(filter_list, list) else [filter_list]
+            filter_list = (
+                filter_list if isinstance(filter_list, list) else [filter_list]
+            )
         elif f"@{filter_key}" in raw_f:
             # perform an attribute filter on each host
             filter_list = list(
                 filter(
-                    lambda item: eval(
+                    lambda item: eval(  # noqa: S307
```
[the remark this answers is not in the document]
JacobCallahan: unfortunately, no. we need to pass in the filterable data structure and ast.literal_eval can't do that.

broker/helpers.py
```diff
@@ -494,7 +526,7 @@ def find_origin():
             return f"{frame.function}:{frame.filename}", jenkins_url
         if frame.function == "call_fixture_func":
             # attempt to find the test name from the fixture's request object
-            if request := _frame.frame.f_locals.get("request"):
+            if request := _frame.frame.f_locals.get("request"):  # noqa: F821
```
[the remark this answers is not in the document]
JacobCallahan: This is largely due to the order in which the frames are available, so it gives us the ability to "look back" at the previous frame to get the request object from the actual fixture instead of the pytest-internal frame we'd be at otherwise.

broker/providers/container.py
```diff
@@ -101,26 +107,26 @@ def __init__(self, **kwargs):
...
+        """Check if an image exists on the provider, attempt a pull if not."""
         for image in self.runtime.images:
             if name in image.tags:
                 return
             elif ("localhost/" in name) and (name[10:] in image.tags):
                 return
         try:
             self.runtime.pull_image(name)
-        except Exception as err:
+        except Exception as err:  # noqa: BLE001 - This could be a few things
```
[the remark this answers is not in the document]
JacobCallahan: it could be, but at this point I'm not sure what all we could hit here.

ruff.toml
JacobCallahan: We don't really use section headings and I think it's ugly.

ruff.toml
JacobCallahan: Many times the contents of an __init__ docstring is a repetition from the class' own docstring. I find docstrings like "Initialize the class" to be useless. If I could turn it off for all other common dunder methods, I would.

ruff.toml
JacobCallahan: See comment about D407 ;)

ruff.toml
JacobCallahan: Sane is up to personal preference and tool compatibility. Most of the times this was triggered in Broker was due to click command options.

ruff.toml
JacobCallahan: oh good catch. looks like it's fixed in 0.0.277

ruff.toml
JacobCallahan: We can use D202, but I don't like the others.

ruff.toml
JacobCallahan: oh definitely, they were the standard for most of this file! however, let's go ahead and remove this ;)

JacobCallahan:
@ogajduse thanks for the thorough review! Updated all I saw to address and a little more.

broker/logger.py
```diff
@@ -21,29 +21,28 @@ class LOG_LEVEL(IntEnum):
...
         return True
 
     def redact_dynaconf(self, data):
-        """
-        This method goes over the data and redacts all values of keys
-        that match the sensitive ones
-        """
-        if isinstance(data, (list, tuple)):
+        """Go over the data and redact all values of keys that match the sensitive ones."""
+        if isinstance(data, list | tuple):
```
JacobCallahan, line 40: This change is incompatible with Python 3.9. Which brings up a question: is it too soon to drop 3.9 support with Broker 0.4? We do have Python 3.12 coming out later this year, so it won't be much longer, but maybe now is too soon..

Griffin-Sullivan: I would say go ahead and keep it. If anyone is using 3.9 they shouldn't...

ruff.toml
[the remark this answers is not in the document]
JacobCallahan: Actually... newest change switched us over to pyproject.toml completely

ogajduse (approved):

broker/broker.py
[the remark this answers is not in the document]
JacobCallahan: I'll keep as-is since the creation of the exception class below will also log as an error.

PR #227, opened 2023-08-21
Updated catalog-info.yaml file for Backstage

rujutashinde:
Modified the catalog-info.yaml file to add namespace, owner and lifecycle

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -21,5 +21,5 @@ metadata:
     github.com/project-slug: SatelliteQE/broker
 spec:
   type: library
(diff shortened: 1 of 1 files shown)


rujutashinde:
@JacobCallahan Could you please review this PR. I have updated it so that we can view it on console.one backstage instance

JacobCallahan (approved):

rujutashinde:
@JacobCallahan So the checks have failed, but it looks like an issue not related to the catlog changes, could we still merge this or the checks need to be fixed first?

PR #241, opened 2023-09-01
Corrected catalog file with tech docs and rover group

rujutashinde:


rujutashinde:
@JacobCallahan can you PTAL

JacobCallahan (approved):

Issue #246, opened 2023-09-19
Feature request: Support authentication via ssh-agent


dosas:
[ssh2-python](https://ssh2-python.readthedocs.io/en/latest/session.html#ssh2.session.Session.agent_auth) supports authentication using ssh-agent.

Could this be added [here](https://github.com/SatelliteQE/broker/blob/master/broker/session.py#L61) additionally to password and key file?

There are setups where for security reasons the password is not given and only login via ssh-key is allowed. Furthermore the key is stored in the agent as not to be leaked in any CI setup.

dosas:
@JacobCallahan yes that is exaclty what I was thinking about :) Very nice.

I came accross this project when trying to run some tests with robottelo.

JacobCallahan:
Ahh nice, welcome!

dosas:
Should I create a PR?

JacobCallahan:
@dosas If you're willing to propose the change and run through a few test scenarios, then please do!

dosas:
I could not find any tests for session, what do you mean by run through a few test scenarios?

JacobCallahan:
@dosas since you can't really run our functional tests (no access to internal resources), then that would be manual tests using the different authentication methods against a remote host of your own. Of course, please sanitize any sensitive information.

```python
# key-based auth
test_host = Host(hostname="my.test.host", key_filename="/path/to/key")
assert test_host.hostname in test_host.execute("hostname").stdout


# password-based auth
test_host = Host(hostname="my.test.host", username="myuser", password="mypassword")
assert test_host.hostname in test_host.execute("hostname").stdout


# agent-based auth
test_host = Host(hostname="my.test.host", username="myuser")
assert test_host.hostname in test_host.execute("hostname").stdout
```

PR #248, opened 2023-09-26
Authentication using ssh-agent

dosas:
This PR adds authentication using ssh-agent, see https://github.com/SatelliteQE/broker/issues/246

Also exception handling for authentication is unified.

### DIFF
--- a/broker/session.py
+++ b/broker/session.py
@@ -54,6 +54,7 @@ class Session:
         sock.settimeout(kwargs.get("timeout"))
         port = kwargs.get("port", 22)
         key_filename = kwargs.get("key_filename")
+        password = kwargs.get("password")
         timeout = kwargs.get("timeout", 60)
         helpers.simple_retry(sock.connect, [(host, port)], max_timeout=timeout)
         self.session = ssh2_Session()
@@ -61,9 +62,15 @@ class Session:
         if key_filename:
             if not Path(key_filename).exists():
                 raise FileNotFoundError(f"Key not found in '{key_filename}'")
-            self.session.userauth_publickey_fromfile(user, key_filename)
-        elif kwargs.get("password"):
-            self.session.userauth_password(user, kwargs["password"])
+            try:
+                self.session.userauth_publickey_fromfile(user, key_filename)
+            except Exception as err:
+                raise exceptions.AuthenticationError("Key-based authentication failed.") from err
+        elif password:
+            try:
+                self.session.userauth_password(user, password)
+            except Exception as err:
+                raise exceptions.AuthenticationError("Password-based authentication failed.") from err
         elif user:
             try:
                 self.session.agent_auth(user)
--- a/broker/session.py
+++ b/broker/session.py
@@ -64,6 +64,11 @@ class Session:
             self.session.userauth_publickey_fromfile(user, key_filename)
         elif kwargs.get("password"):
             self.session.userauth_password(user, kwargs["password"])
+        elif user:
+            try:
+                self.session.agent_auth(user)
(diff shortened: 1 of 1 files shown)


broker/session.py
```diff
@@ -54,16 +54,28 @@ def __init__(self, **kwargs):
...
+                self.session.userauth_password(user, password)
+            except Exception as err:
+                raise exceptions.AuthenticationError("Password-based authentication failed.") from err
+        elif user:
+            try:
+                self.session.agent_auth(user)
+            except Exception as err:
+                raise exceptions.AuthenticationError("Agent-based authentication failed.") from err
         else:
             raise exceptions.AuthenticationError("No password or key file provided.")
```
JacobCallahan, line 80: I like where you're going with, but it looks a bit more repetitive than I'd prefer. Perhaps we could wrap the entire section in a try/except and raise a formatted message like this.
```python
        try:
            if key_filename:
                auth_type = "Key"
                if not Path(key_filename).exists():
                    raise FileNotFoundError(f"Key not found in '{key_filename}'")
                self.session.userauth_publickey_fromfile(user, key_filename)
            elif password:
                auth_type = "Password"
                self.session.userauth_password(user, password)
            elif user:
                auth_type = "Agent"
                self.session.agent_auth(user)
            else:
                raise exceptions.AuthenticationError("No password or key file provided.")
        except Exception as err:
            raise exceptions.AuthenticationError(f"{auth_type}-based authentication failed.") from err
```

JacobCallahan (changes requested):
Also, please try installing pre-commit in your virtual environment and run `pre-commit` when you stage changes. You can also run it against specific files or `--all-files`.
This will get your checks passing!

broker/session.py
JacobCallahan: `auth_type` isn't used in the else clause, so shouldn't cause an issue.
as for ruff, you can add this to the right of the except statement `# noqa: BLE001` there are likely a few different exceptions that could be raised here and I'd rather not waste your time chasing them all down.

JacobCallahan (changes requested):
Can you provide a demonstration of agent/session-based authentication working for you locally?

dosas:
I am not exactly sure what I should put here but I have written some tests and ran them locally:

```
def test_key():
    test_host = Host(hostname=hostname, key_filename="id_ed25519")
    assert "host1" in test_host.execute("hostname").stdout, "path"


def test_wrong_key():
    test_host = Host(hostname=hostname, key_filename="wrong_key")
    with pytest.raises(BrokerAuthenticationError):
        test_host.execute("hostname").stdout

def test_password():
    test_host = Host(hostname=hostname, username="root", password="bar")
    assert "host1" in test_host.execute("hostname").stdout, "password"


def test_wrong_password():
    test_host = Host(hostname=hostname, username="root", password="foo")
    with pytest.raises(BrokerAuthenticationError):
        test_host.execute("hostname").stdout


def test_agent():
    test_host = Host(hostname=hostname, username="root")
    assert "host1" in test_host.execute("hostname").stdout, "agent"

```

with the wollowing results:

```
test_manual.py::test_key PASSED                                                                                                                                                        [ 20%]
test_manual.py::test_wrong_key PASSED                                                                                                                                                  [ 40%]
test_manual.py::test_password PASSED                                                                                                                                                   [ 60%]
test_manual.py::test_wrong_password PASSED                                                                                                                                             [ 80%]
test_manual.py::test_agent PASSED
```

JacobCallahan:
Awesome, thanks for the contribution @dosas!

JacobCallahan (approved):

PR #254, opened 2023-12-05
Included catalog-info.yaml for the broker repository with the project informaion.

riginoommen:
In this PR we have included the catalog-info file which is needed for the backstage to identify and parse the information related to the broker

PR #256, opened 2023-12-13
Added system and location information for catalog-info.yaml

rujutashinde:
Added system and location information so the Broker tool can be seen correctly under quality-community domain

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -1,26 +1,12 @@
-# Catalog entry for Backstage [backstage.io]
 apiVersion: backstage.io/v1alpha1
-kind: Component
+kind: Location
 metadata:
-  name: broker
-  title: Broker
-  description: The infrastructure middleman!
-  links:
-    - title: PyPi
-      url: https://pypi.org/project/broker
-  tags:
-    - python
-    - infrastructure
-    - click
-    - ansible-tower
-    - aap
-    - docker
-    - podman
-  namespace: quality-community   
-  annotations:
-    github.com/project-slug: SatelliteQE/broker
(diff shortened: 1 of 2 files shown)


rujutashinde:
@JacobCallahan I put in a new PR instead of PR #254  we can close that one, so we are not waiting on updates there

catalog-info.yaml
```diff
@@ -1,26 +1,11 @@
...
+  title: QE Broker Root Configuration
+  namespace: quality-community
+  description: A root collection of all QE Broker configurations for Backstage.
 spec:
-  type: library
-  owner: group:redhat/satellite-qe
-  lifecycle: preproduction
+  owner: group:redhat/quality-community
+  targets:
+    - ./catalog/system/broker.yaml
```
JacobCallahan, line 11:
```suggestion
    - ./catalog/system/broker.yaml

```

catalog/system/broker.yaml
```diff
@@ -0,0 +1,24 @@
...
+    - ansible-tower
+    - aap
+    - docker
+    - podman
+  links:
+    - title: PyPi
+      url: https://pypi.org/project/broker
+spec:
+  owner: group:redhat/satellite-qe
+  domain: quality-community/quality-community
```
JacobCallahan, line 24:
```suggestion
  domain: quality-community/quality-community

```

JacobCallahan (changes requested):
Missing some trailing newlines. Also, can you update your commit message to describe the changes, specifically why the additional file is used for metadata instead of the original?

JacobCallahan (approved):

PR #260, opened 2024-01-29
Add IPv6 support

JacobCallahan:
This change is mostly focused around the addition of IPv6 connection functionality to Broker's Host Sessions.
Included are options for an ipv6 connection preference with an opt-out ipv6 connection fallback.
These have also been added to the settings file.

### DIFF
--- a/broker/hosts.py
+++ b/broker/hosts.py
@@ -35,13 +35,15 @@ class Host:
         """Create a Host instance.
 
         Expected kwargs:
-          hostname: str - Hostname or IP address of the host, required
-          name: str - Name of the host
-          username: str - Username to use for SSH connection
-          password: str - Password to use for SSH connection
-          connection_timeout: int - Timeout for SSH connection
-          port: int - Port to use for SSH connection
-          key_filename: str - Path to SSH key file to use for SSH connection
+            hostname: (str) - Hostname or IP address of the host, required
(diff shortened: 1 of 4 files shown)


broker/hosts.py
```diff
@@ -35,13 +35,15 @@ def __init__(self, **kwargs):
...
-          key_filename: str - Path to SSH key file to use for SSH connection
+            hostname: str - Hostname or IP address of the host, required
+            name: str - Name of the host
+            username: str - Username to use for SSH connection
+            password: str - Password to use for SSH connection
+            connection_timeout: int - Timeout for SSH connection
+            port: int - Port to use for SSH connection
+            key_filename: str - Path to SSH key file to use for SSH connection
+            ipv6 (bool): Whether or not to use IPv6. Defaults to False.
+            ipv4_fallback (bool): Whether or not to fallback to IPv4 if IPv6 fails. Defaults to True.
```
jyejare, line 46: Should we keep the `type` format in docstrings same across ?

jyejare (approved):

broker/hosts.py
JacobCallahan: yeah, i can make that update

jyejare:
@JacobCallahan Verified the combination of IPV6_HOST set to True and FALLBACK to False , it worked as expected!

PR #261, opened 2024-01-30
Use host facts when syncing inventory from AAP

tpapaioa:
Get **_broker_args** from host's **ansible_facts**, if present, instead of from the job / workflow job attributes.

### DIFF
--- a/broker/providers/ansible_tower.py
+++ b/broker/providers/ansible_tower.py
@@ -364,15 +364,15 @@ class AnsibleTower(Provider):
             "_broker_provider": "AnsibleTower",
             "_broker_provider_instance": self.instance,
             # Get _broker_args from host facts if present
-            "_broker_args": getattr(host_facts, "_broker_args", {}) or self._get_broker_args_from_job(host),
+            "_broker_args": getattr(
+                host_facts, "_broker_args", self._get_broker_args_from_job(host)
+            ),
         }
 
         return host_info
 
     def _get_broker_args_from_job(self, host):
-        """Get _broker_args from the source workflow job.
-           If the source workflow job could not be found, then try the last job.
-        """
+        """Get _broker_args from the source workflow job or last job."""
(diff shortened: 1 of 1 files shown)


broker/providers/ansible_tower.py
```diff
@@ -366,36 +372,42 @@ def _compile_host_info(self, host):
             "tower_inventory": self._translate_inventory(host.inventory),
             "_broker_provider": "AnsibleTower",
             "_broker_provider_instance": self.instance,
-            "_broker_args": getattr(host, "_broker_args", {}),
+            # Get _broker_args from host facts if present
+            "_broker_args": getattr(host_facts, "_broker_args", {}) or self._get_broker_args_from_job(host),
```
JacobCallahan, line 376: since this one is just the single fallback, can just have it be the default for get.
```suggestion
            "_broker_args": getattr(host_facts, "_broker_args", self._get_broker_args_from_job(host)),
```

broker/providers/ansible_tower.py
```diff
@@ -434,6 +446,11 @@ def construct_host(self, provider_params, host_classes, **kwargs):
                 if "tower_inventory" in job_attrs
                 else self._translate_inventory(job.summary_fields.inventory)
             }
+
+            # FIXME Find better way to return ip / ansible_host from job
+            misc_attrs["ip"] = kwargs["ip"]
+            kwargs.pop("ip")
```
JacobCallahan, line 452: can do it on a single line with a safety if the ip isn't there
```suggestion
            misc_attrs["ip"] = kwargs.pop("ip", None)
```

JacobCallahan (changes requested):
Minor changes. Also, please run pre-commit before pushing back up

JacobCallahan (approved):
Checks are passing and the changes proposed look good. Aside from a new merge conflict from #262 is this ready to merge?

PR #269, opened 2024-02-15
truncate existing files on sftp write

rmynar:
In our use case we create a backup of a config file, modify the config file (adding characters), perform testing and restore the original config from backup. We use sftp to read and write the backup. However the restore (sftp_write) ends with damaged config file. Damaged file consists of original content and few characters remained from modified file.

Adding flag LIBSSH2_FXF_TRUNC should solve the problem by truncating existing files.
see: https://libssh2.org/libssh2_sftp_open_ex.html

### DIFF
--- a/broker/session.py
+++ b/broker/session.py
@@ -27,7 +27,9 @@ try:
         | ssh2_sftp.LIBSSH2_SFTP_S_IRGRP
         | ssh2_sftp.LIBSSH2_SFTP_S_IROTH
     )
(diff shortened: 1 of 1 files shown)


lpramuk (approved):
Clear file before writing will resolve issue with successive writes to the same file.
Good catch @rmynar !

ogajduse (approved):
ACK, good find, @rmynar.

Only one thing comes to my mind...
@JacobCallahan Would you like to preserve the current behavior by allowing the user to decide whether to truncate the file?

JacobCallahan:
@ogajduse that's a good suggestion to keep in mind for the future if needed, but would complicate the current design a bit. Let's hold off on that for now and revisit if issues arise from this.

PR #271, opened 2024-03-01
Vendor ssh2-python as a submodule

JacobCallahan:
In another attempt to circumvent the restrictions from PyPI, I am adding a submodule for my fork of ssh2-python and adding that as a local build requirement.


ogajduse (approved):

jyejare:
@JacobCallahan We also need to update the broker req in robottelo master!

JacobCallahan:
@jyejare not yet. I'm actually going to be reverting most of this because pip doesn't like it. Unfortunately, this was a failed experiment.

PR #275, opened 2024-03-13
Swap ssh2-python dependency to ssh2-python312

JacobCallahan:
This change moves away from the github fork-based dependency to the newly published ssh2-python312 on PyPI.


JacobCallahan:
The functional test `test_tower_host_mp`, which tests most ssh functionality we use, has passed locally.

PR #277, opened 2024-03-18
Add optional provider label support

rplevka:
Extend broker's AnsibleTower and Container provider functionality to pass labels to be set to the spawned resources.
It accepts a string of comma-separated label values.
Broker then uses awxkit session to resolve their id's (required for the job `/launch` payload) and eventually creates the non-existing labels.

User is required to have granted permissions for creating new labels (if that's even a thing) on AAP.

Plan is to implement the labeling functionality to all providers that have support of labels, e.g.
https://docs.docker.com/config/labels-custom-metadata/

example usage:
```
# ansible tower provider
$ broker execute deploy-rhel --provider-labels foo=bar,baz
```
translates to:
Selection_674
(AAP labels are single string items)

```
# container provider with additional labels configured in the nick:

nick:

  rhel8_docker:
    container_host: "ubi8:latest"
    environment:
      BUILD_URL: "https://intergalactic.planetary/planetary/intergalactic"
    provider_labels:
      nick_label_1: myval1
      nick_label_2: myval2


$ broker checkout --nick rhel8_docker --provider-labls cli1=val1,nick_label_2=overriden_by_cli_param
[INFO 240418 14:13:37] Using provider Container to checkout
[INFO 240418 14:13:45] Host: 97c993636de8
```
translates to:
```
$ docker inspect 97c993636de8
...
            "Labels": {
                "architecture": "x86_64",
                "broker.jenkins.url": "",
                "broker.nick_label_1": "myval1",
                "broker.nick_label_2": "myval2",
                "broker.origin": "broker_cli:rplevka",
               ...
...
```

### DIFF
--- a/broker/commands.py
+++ b/broker/commands.py
@@ -191,7 +191,7 @@ def cli(version):
 @click.option("-c", "--count", type=int, help="Number of times broker repeats the checkout")
 @click.option(
     "-l",
-    "--labels",
+    "--provider-labels",
     type=str,
     help="A string representing the list"
     " of k=v pairs (comma-separated) to be used as provider resource"
@@ -204,7 +204,7 @@ def cli(version):
 )
 @provider_options
 @click.pass_context
-def checkout(ctx, background, nick, count, args_file, labels, **kwargs):
+def checkout(ctx, background, nick, count, args_file, provider_labels, **kwargs):
     """Checkout or "create" a Virtual Machine broker instance.
 
     COMMAND: broker checkout --workflow "workflow-name" --workflow_arg1 something
@@ -218,10 +218,10 @@ def checkout(ctx, background, nick, count, args_file, labels, **kwargs):
         broker_args["_count"] = count
     if args_file:
         broker_args["args_file"] = args_file
-    if labels:
-        broker_args["labels"] = {
-            f"broker.{label[0]}": "=".join(label[1:])
-            for label in [kv_pair.split("=") for kv_pair in labels.split(",")]
+    if provider_labels:
+        broker_args["provider_labels"] = {
+            label[0]: "=".join(label[1:])
+            for label in [kv_pair.split("=") for kv_pair in provider_labels.split(",")]
         }
 
     # if additional arguments were passed, include them in the broker args
--- a/broker/commands.py
+++ b/broker/commands.py
@@ -190,8 +190,13 @@ def cli(version):
 @click.option("-n", "--nick", type=str, help="Use a nickname defined in your settings")
 @click.option("-c", "--count", type=int, help="Number of times broker repeats the checkout")
 @click.option(
-    "--labels", type=str, help="List of strings (comma-separated) to be used as AAP labels"
-)  # fixme - this should be provider-agnostic
+    "-l",
+    "--labels",
+    type=str,
+    help="A string representing the list"
+    " of k=v pairs (comma-separated) to be used as provider resource"
+    " labels (e.g. '-l k1=v1,k2=v2,k3=v3=z4').",
+)
 @click.option(
     "--args-file",
     type=click.Path(exists=True),
--- a/broker/commands.py
+++ b/broker/commands.py
@@ -190,13 +190,16 @@ def cli(version):
 @click.option("-n", "--nick", type=str, help="Use a nickname defined in your settings")
 @click.option("-c", "--count", type=int, help="Number of times broker repeats the checkout")
 @click.option(
+    "--labels", type=str, help="List of strings (comma-separated) to be used as AAP labels"
+)  # fixme - this should be provider-agnostic
+@click.option(
     "--args-file",
     type=click.Path(exists=True),
     help="A json or yaml file mapping arguments to values",
 )
 @provider_options
 @click.pass_context
-def checkout(ctx, background, nick, count, args_file, **kwargs):
+def checkout(ctx, background, nick, count, args_file, labels, **kwargs):
     """Checkout or "create" a Virtual Machine broker instance.
 
     COMMAND: broker checkout --workflow "workflow-name" --workflow_arg1 something
@@ -210,6 +213,8 @@ def checkout(ctx, background, nick, count, args_file, **kwargs):
         broker_args["_count"] = count
     if args_file:
         broker_args["args_file"] = args_file
+    if labels:
+        broker_args["labels"] = labels.split(",")
     # if additional arguments were passed, include them in the broker args
     # strip leading -- characters
     broker_args.update(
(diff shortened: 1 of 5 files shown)


broker/providers/ansible_tower.py
```diff
@@ -572,6 +591,12 @@ def execute(self, **kwargs):  # noqa: PLR0912 - Possible TODO refactor
         if inventory := kwargs.pop("inventory", None):
             payload["inventory"] = inventory
             logger.info(f"Using tower inventory: {self._translate_inventory(inventory)}")
+        if labels := kwargs.pop("provider_labels", None):
+            payload["labels"] = self._resolve_labels(labels, target)
+            # record labels also as extra vars - use key=value format
+            kwargs.update(
+                {f"_broker_label_{label[0]}": "=".join(label[1:]) for label in labels.items()}
```
JacobCallahan, line 598: each label getting its own attribute may be a bit much. what do you think about collecting these under a `_provider_labels` list? or maybe even just `labels`?

JacobCallahan (changes requested):
Can you add a functional test each for tower and containers? 
I think we can just skip beaker for now since I'm not sure that labels are a thing there (and it is much less commonly used).

Also, how well are these labels recovered after a sync?

broker/commands.py
```diff
@@ -187,14 +195,22 @@ def cli(version):
 @click.option("-b", "--background", is_flag=True, help="Run checkout in the background")
 @click.option("-n", "--nick", type=str, help="Use a nickname defined in your settings")
 @click.option("-c", "--count", type=int, help="Number of times broker repeats the checkout")
+@click.option(
+    "-l",
+    "--provider-labels",
+    type=str,
+    help="A string representing the list"
+    " of k=v pairs (comma-separated) to be used as provider resource"
+    " labels (e.g. '-l k1=v1,k2=v2,k3=v3=z4').",
```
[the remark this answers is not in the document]
rplevka: i somehow wanted to demo, that multiple `=` won't be an issue (translates to `k3: `v3=z4`) but apparently I ended up confusing you :)

broker/commands.py
```diff
@@ -289,13 +308,23 @@ def inventory(details, sync, filter):
...
+)
 @provider_options
-def extend(vm, background, all_, sequential, filter, **kwargs):
+def extend(vm, provider_labels, background, all_, sequential, filter, **kwargs):
```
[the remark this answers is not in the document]
rplevka: ...well, I am happy to drop this, but on the other hand - it generates an AAP job. And perhaps it deserves to have the labels assigned too. I doubt we would use it too much tho

broker/providers/ansible_tower.py
rplevka: This would require additional parsing logic to be implemented in AT (another round of `.split('=')`). What are the downsides of passing them as extra vars straight away?

broker/commands.py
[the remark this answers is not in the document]
bherrin3: the main place I care about labels is when we are creating or destroying resources in providers. This causes a need for load and resource cost. 

Since extend is purely an SLA thing, I would state labels are not required here, for now.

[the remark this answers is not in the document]
JacobCallahan: cool. let's remove here then

broker/providers/ansible_tower.py
```diff
@@ -436,6 +436,32 @@ def _pull_extra_vars(extra_vars):
...
+        """
+        label_ids = []
+        for label in labels:
+            label_expanded = f"{label}={labels[label]}" if labels[label] else label
+            try:
+                if result := self.v2.labels.get(name=label_expanded).results:
+                    label_ids.append(result[0].id)
+                else:
+                    # label does not exist yet, creating
+                    result = self.v2.labels.post(
```
JacobCallahan, line 452: do we expect both of these to possibly give us a permission error or just the label creation? i typically like to isolate try/except to a single suspect endpoint.

broker/providers/ansible_tower.py
```diff
@@ -436,6 +436,32 @@ def _pull_extra_vars(extra_vars):
...
+                        {
+                            "name": label_expanded,
+                            "organization": target.summary_fields.organization.id,
+                        }
+                    )
+                    if result:
+                        label_ids.append(result.id)
+            except awxkit.exceptions.Forbidden:
+                raise exceptions.PermissionError
```
JacobCallahan, line 461: can you add a message here telling the user that the permission error is because they were unable to read/create the label?

tests/functional/test_satlab.py
```diff
@@ -122,3 +122,21 @@ def test_tower_host_mp():
...
+        assert r_host.provider_labels.get("l1") == "v1"
+        assert r_host.provider_labels.get("l2") == ""
+        # assert the AAP labels got created on the provider
+        aap_labels = [l.name for l in r_host._prov_inst.v2.labels.get().results]
+        assert "l1=v1" in aap_labels
+        assert "l2" in aap_labels
+
+def test_fero(mocker):
+    with Broker(workflow="deploy-rhel", provider_labels={"l1": "v1", "l2": ""}) as r_host:
+        mocker.patch.object()
```
JacobCallahan, line 142: why the use of pytest-mock here? also, there's a missing newline

JacobCallahan (changes requested):
Just a few more things. Can you also resolve the merge conflicts?

broker/providers/ansible_tower.py
rplevka: so, according @bherrin3 , both of these should be "Free" and should not require any extra permissions. any authenticated, regular user, should be able to list and create labels.
Would you prefer to keep the try/except block in place just for a case?

bherrin3: Verified and tested with non-admin users.

Removal of labels is purely an administrative function that tower does itself, but would never happen due to the way we are using labels and not cleaning up jobs.

If listing labels is desired, I would see this as a function of broker for the AnsibleTower provider as that API endpoint literally has the count and all labels in a data structure.

tests/functional/test_satlab.py
rplevka: that was a part of my experiment, not intended to end up here (already removed)

broker/providers/ansible_tower.py
rplevka: done

broker/providers/ansible_tower.py
rplevka: obsoleted by removing the whole try-except block

pyproject.toml
```diff
@@ -42,6 +42,7 @@ beaker = ["beaker-client"]
 dev = [
     "pre-commit",
     "pytest",
+    "pytest-mock",
```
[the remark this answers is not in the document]
rplevka: ah, right. thanks for reminder. removed.

JacobCallahan (approved):

PR #288, opened 2024-04-16
Add full compatibility for podman

JacobCallahan:
This change fills the gap previously seen for podman's inability execute commands on containers.
A large part of this was handling the returned output from podman's exec_run, which is a tuple containing the return code and a bytes string containing both the stdout and stderr.
Due to this, we needed to implement a demuxer.

Also made some minor tweaks to the functional tests

### DIFF
--- a/broker/helpers.py
+++ b/broker/helpers.py
@@ -20,6 +20,7 @@ from logzero import logger
 import yaml
 
 from broker import exceptions, logger as b_log, settings
+from broker.binds.containers import demux_output
 
 FilterTest = namedtuple("FilterTest", "haystack needle test")
 INVENTORY_LOCK = threading.Lock()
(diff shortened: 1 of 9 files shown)


tpapaioa (approved):
LGTM

broker/helpers.py
```diff
@@ -511,8 +512,15 @@ def from_ssh(cls, stdout, channel):
         )
 
     @classmethod
-    def from_duplexed_exec(cls, duplex_exec):
+    def from_duplexed_exec(cls, duplex_exec, runtime=None):
         """Create a Result object from a duplexed exec object from the docker library."""
```
omkarkhatavkar, line 516: just not docker

JacobCallahan: updated, thanks

PR #291, opened 2024-05-03
Foreman broker

knoppi:
As dicussed in https://github.com/SatelliteQE/broker/discussions/284 we created a draft for a foreman broker. Command execution relies on foreman's remote execution.

No configuration of foreman is done by this broker and host provisioning relies solely on correctly configured host groups. The broker can understand the parameters `image` and `compute_resource` to do an image based deployment which accelerates the checkout process.


broker/providers/foreman.py
```diff
@@ -0,0 +1,427 @@
...
+from logzero import logger
+import requests
+
+from broker.exceptions import ProviderError
+from broker.helpers import Result
+from broker.providers import Provider
+from broker.settings import settings
+
+
+class ForemanAPI:
```
JacobCallahan, line 16: Please move this class to a module under the binds directory since it is the abstraction layer between the provider and broker.

broker/providers/foreman.py
```diff
@@ -0,0 +1,427 @@
...
+
+    def __init__(self, **kwargs):
+        self.foreman_username = settings.foreman.foreman_username
+        self.foreman_password = settings.foreman.foreman_password
+        self.url = settings.foreman.foreman_url
+        self.prefix = settings.foreman.name_prefix
+        self.verify = settings.foreman.verify
+        self.session = requests.session()
+
+    def interpret_response(self, response):
```
JacobCallahan, line 31: Since this wouldn't be a "public" method, let's use a leading underscore.
```suggestion
    def _interpret_response(self, response):
```

broker/providers/foreman.py
```diff
@@ -0,0 +1,427 @@
...
+        """Handle responses from Foreman, in particular catch errors."""
+        if "error" in response:
+            if "Unable to authenticate user" in response["error"]["message"]:
+                logger.warning("Could not authenticate")
```
JacobCallahan, line 35: Broker has an AuthenticationError you can use in this case.

broker/providers/foreman.py
```diff
@@ -0,0 +1,427 @@
...
+            raise ProviderError(
+                provider=self.__class__.__name__,
+                message=" ".join(response["error"]["full_messages"]),
+            )
+        if "errors" in response:
+            raise ProviderError(
```
JacobCallahan, line 41: Since this class is going to be moved to a bind, you can create a new ForemanBind exception and raise that here.

broker/providers/foreman.py
```diff
@@ -0,0 +1,427 @@
...
+            **kwargs,
+        )
+        return self.interpret_response(response)
+
+    def obtain_id_from_name(self, resource_type, resource_name):
+        """Obtain id for resource with given name.
+
+        :param resource_type: Resource type, like hostgroups, hosts, ...
+
+        :param resource_name: String-like identifiere of the resource
```
JacobCallahan, line 83: Too French ;)
```suggestion
        :param resource_name: String-like identifier of the resource
```

broker/providers/foreman.py
```diff
@@ -0,0 +1,427 @@
...
+            resource = next(x for x in result if x["name"] == resource_name)
+            id_ = resource["id"]
+        except KeyError:
+            logger.error(f"Could not find {resource_type} {resource_name}")
+            raise
+        except StopIteration:
+            raise ProviderError(
+                provider=self.__class__.__name__,
+                message=f"Could not find {resource_name} in {resource_type}",
+            )
```
JacobCallahan, line 101: Both of these should raise a ForemanBindError

broker/providers/foreman.py
```diff
@@ -0,0 +1,427 @@
...
+        """
+        still_running = True
+        while still_running:
+            response = self._get(f"/api/job_invocations/{job_id}")
+            still_running = response["status_label"] == "running"
+            time.sleep(1)
+
+    def hostgroups(
+        self,
+    ):
```
JacobCallahan, line 128:
```suggestion
    def hostgroups(self):
```

broker/providers/foreman.py
```diff
@@ -0,0 +1,427 @@
...
+    def image_uuid(self, compute_resource_id, image_name):
+        """Return the uuid of a VM image on a specific compute resource."""
+        try:
+            return self._get(
+                "/api/compute_resources/"
+                f"{compute_resource_id}"
+                f"/images/?search=name={image_name}"
+            )["results"][0]["uuid"]
+        except IndexError:
+            logger.error(f"Could not find {image_name} in VM images")
```
JacobCallahan, line 145: is this a recoverable condition, or should an exception be raised?

broker/providers/foreman.py
```diff
@@ -0,0 +1,427 @@
...
+    _sensitive_attrs = ["foreman_password"]
+
+    _checkout_options = [
+        click.option(
+            "--hostgroup",
+            type=str,
+            help="Name of the Foreman hostgroup to deploy the host",
+        ),
+    ]
+    _execute_options = []
```
JacobCallahan, line 184: were you planning on using the job invocation for executes at this time or a later time?

broker/providers/foreman.py
```diff
@@ -0,0 +1,427 @@
...
+        rex=None,
+        **kwargs,
+    ):
+        """Return useful information about Foreman provider."""
+        if hostgroups:
+            self.__init__()
+            all_hostgroups = self.runtime.hostgroups()
+            logger.info(f"On Foreman {self.instance} you have the following hostgroups:")
+            for hostgroup in all_hostgroups["results"]:
+                logger.info(f"- {hostgroup['title']}")
```
JacobCallahan, line 309: It would be nice to also get information about a hostgroup here as well, so a user could pass `broker providers Foreman --hostgroup myhostgroup` and see its details.

broker/providers/foreman.py
```diff
@@ -0,0 +1,427 @@
...
+
+    def _compile_host_info(self, host):
+        return {
+            "name": host["certname"],  # alternatives: name, display_name
```
JacobCallahan, line 313: not recommending a change here, but why `certname` over `display_name` or more simply `name`?

JacobCallahan (changes requested):
Overall great job with this addition, I'm very impressed!
I mostly just have some organizational requests.

broker/providers/foreman.py
knoppi: Done that

broker/providers/foreman.py
knoppi: Done that

broker/providers/foreman.py
knoppi: Okay, I switched to using AuthenticationError.

broker/providers/foreman.py
knoppi: Perfect, didn't know that - step by step I get the overall structure.

broker/providers/foreman.py
knoppi: Fixed

broker/providers/foreman.py
knoppi: Fixed

broker/providers/foreman.py
knoppi: Technically, we could implement a fallback to network-based deployment or whatever the default is for the hostgroup, but I think raising an Exception is the desired option here. So, I changed in that direction.

broker/providers/foreman.py
knoppi: It is implemented to use job invocation for executes and I already tried it within robottelo tests (mediated through the created host object).

I had no use for implementing this as a standalone feature for the command line so far. Or am I missing something here?

broker/providers/foreman.py
knoppi: I've implemented this in the new commits. But I'm not sure to which details should be displayed. I chose a minimal set.

broker/providers/foreman.py
knoppi: Good point. I assumed certname to always contain the FQDN where name could be the shortname. Hence, I tried to always point to a FQDN, since this is also what is required in robottelo tests.

broker/providers/foreman.py
JacobCallahan: I have no solid requirement for hooking up the execute functionality through the cli and Broker's execute method. I was primarily just thinking that it looks like you have everything needed here to set that up.

broker/providers/foreman.py
```diff
@@ -0,0 +1,296 @@
...
+
+    def provider_help(
+        self,
+        hostgroups=False,
+        hostgroup=None,
+        **kwargs,
+    ):
+        """Return useful information about Foreman provider."""
+        if hostgroups:
+            self.__init__()
```
JacobCallahan, line 157: What is the need for running the init again?

JacobCallahan (approved):
Thanks for the quick updates to this PR and the great effort overall on it!

JacobCallahan:
@knoppi It looks like your test failure is due to a chicken/egg design limitation around Broker's config that I can likely address for 0.5. I assume these tests are passing locally.

knoppi:
> @knoppi It looks like your test failure is due to a chicken/egg design limitation around Broker's config that I can likely address for 0.5. I assume these tests are passing locally.

There might be some design issue with my test as well. They were passing locally when I ran them with a non-working, but valid config. Yesterday when reworking I realized they fail without pointing to my test settings. They fail during validation because they cannot find the required settings (organization, location, username, password).

I then checked how you do this for other providers. For containers there is no setting defined as `must_exist` and you always have a default to fall-back to. For Ansible Tower the constructor never gets called.

My preferred solution would be reading the settings in the class ForemanBind where they are actually needed, so I could overwrite the validators in the MockStub. With the current design, I see two options:

1. Get rid of tests that rely on the constructor.
2. Store my test settings as defaults within the validators.

Option 1 would leave me rather useless tests. Option 2 seems error-prone but error handling should be sufficient to indicate if you are using the wrong values or forgot to set some.

Do you have an oppinion on that?

broker/providers/foreman.py
knoppi: Good catch. That was an artifact from the beginning.

JacobCallahan:
Since you're loading the settings in the init phase, you also have the option to pass a MockStub version of the settings that has the values you want for your test.
```python
    ...
    settings = kwargs.pop("stub_settings", settings)
```
However, since most of your validators are set to must_exist, you'd still fail dynaconf's settings validation. Hold tight for now and we'll see if the change I'm working on for local settings init will help.

JacobCallahan:
@knoppi I've proposed #293 that will hopefully solve this issue in GitHub actions. If you don't mind checking it over, I can merge it and then a rebase of this PR will test if it works as expected for this.

JacobCallahan:
@knoppi I've merged #293, so please rebase this PR and push back up. I'll then kick off the actions and see how the tests work.

JacobCallahan (approved):
ACK. Thanks for following this all through and adding the new provider!

PR #296, opened 2024-06-16
fixes #295 - handle Duplicate error while creating aap job labels

### LINKED ISSUE #295: provider_labels/aap - label creation suspectible to race conditions
The part behind checking for an eventual existing label and the consecutive POST request is suspectible to a race condition - if in between these 2 calls the label appears on AAP, the POST request will fail with `400` and the corresponding message.

https://github.com/SatelliteQE/broker/blob/master/broker/providers/ansible_tower.py#L460-L466

It would probably be wiser to replace this by a single, `POST` request wrapped inside a `try-except` block and handle the scenario (maybe log a warning or a debug message and pass)

rplevka:
Try to `POST` to job labels straight away and catch eventual `Duplicate` `400` response.
in such case, retrieve the label id using the `GET` method and carry on.

2 consecutive broker checkout commands with passing the same label should be enough to test this.
the first one creates the label on AAP, while the other one hits the `400` and handles it.
As a result, both resulting AAP jobs have the same job label attached and both broker commands finish successfully.

### DIFF
--- a/broker/providers/ansible_tower.py
+++ b/broker/providers/ansible_tower.py
@@ -457,15 +457,21 @@ class AnsibleTower(Provider):
         label_ids = []
         for label in labels:
             label_expanded = f"{label}={labels[label]}" if labels[label] else label
-            if result := self.v2.labels.get(name=label_expanded).results:
-                label_ids.append(result[0].id)
-            else:
-                # label does not exist yet, creating
+            try:
                 result = self.v2.labels.post(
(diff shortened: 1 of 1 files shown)


JacobCallahan (approved):
I like the distinction in your log messages. ACK

ogajduse (approved):

JacobCallahan:
@rplevka is this ready to merge?

rplevka:
@JacobCallahan yes

PR #299, opened 2024-06-27
Enhance port mapping logic, add exposed ports to Host object

ogajduse:
This PR ensures that broker
1. respects exposed ports in/from the image and then extends them by ports provided in arguments.
2. If port 22 was exposed, changes the hostname of the Host to `the.docker.host.com:<exposed-ssh-port>.
3. Adds `exposed_ports` attribute to the Host class if any ports were exposed. Format `{"inner-port":"exposed-port"}` e.g. `{'22': '44263'}`


broker/providers/container.py
```diff
@@ -200,13 +203,18 @@ def construct_host(self, provider_params, host_classes, **kwargs):
...
         if not hostname:
             raise Exception(f"Could not determine container hostname:\n{cont_attrs}")
         name = cont_attrs["name"]
         logger.debug(f"hostname: {hostname}, name: {name}, host type: host")
         host_inst = host_classes["host"](**{**kwargs, "hostname": hostname, "name": name})
         self._set_attributes(host_inst, broker_args=kwargs, cont_inst=cont_inst)
+        # add the container's port mapping to the host instance only if there are any ports open
+        if cont_attrs["ports"]:
+            host_inst.exposed_ports = {
+                f"{k.split('/')[0]}": v[0]["HostPort"] for k, v in cont_attrs["ports"].items()
```
JacobCallahan, line 216: it seems like this may need some validation before being accessed since it's raising an error in the tests
```python
v[0]["HostPort"]
```

JacobCallahan (changes requested):

broker/providers/container.py
ogajduse: After I played with it a bit, I realized that in tests, `cont_attrs` is `ContainerApiStub` instead of a dictionary. Therefore using the `get` method here seems to be better than subscribing directly (calling __getitem__).

I rely on podman-py to provide the data in the following structure `# {'22/tcp': [{'HostIp': '', 'HostPort': '1337'}]`. Would you like to validate the structure anyway?

```python
(Pdb++) type(cont_attrs)
<class 'test_container.ContainerApiStub'>
(Pdb++) cont_attrs
{'images': [{'tags': 'ch-d:ubi8'}], 'containers': [{'tags': 'f37d3058317f'}], 'name': 'f37d3058317f'}
(Pdb++) cont_attrs["ports"]
{'images': [{'tags': 'ch-d:ubi8'}], 'containers': [{'tags': 'f37d3058317f'}], 'name': 'f37d3058317f'}
(Pdb++) cont_attrs.get("ports", False)
False
```

ogajduse: I also added `ports` to `ContainerApiStub`.

JacobCallahan (approved):

PR #302, opened 2024-07-29
Broker changes for remote podman (tcp and ipv6)

JacobCallahan:
We have encountered a couple of issues that need to be solved in order to use Broker with remote podman hosts.

  First is an issue with the ssh-based connection to remote podman that was investigated in a previous issue. The
  solution to this from Broker's side is largely changing the connection string based on the host_port given.

  Second is to give Broker the ability to automatically find and pass along the correct network information when
  spinning up an ipv6 container. The solution for this is to list networks, search for one that is ipv6 enabled, then
  modify the container creation arguments to pass along that network information

### DIFF
--- a/broker/binds/containers.py
+++ b/broker/binds/containers.py
@@ -1,8 +1,11 @@
 """A collection of classes to ease interaction with Docker and Podman libraries."""
+from broker.exceptions import UserError
+from broker.settings import settings
 
 HEADER_SIZE = 8
 STDOUT = 1
 STDERR = 2
+SSH_PORT = 22
 
 
 def demux_output(data_bytes):
@@ -79,6 +82,11 @@ class ContainerBind:
         """Return a list of containers on the container host."""
         return self.client.containers.list(all=True)
 
+    @property
+    def networks(self):
+        """Return a list of networks on the container host."""
+        return self.client.networks.list()
+
     def image_info(self, name):
         """Return curated information about an image on the container host."""
         if image := self.client.images.get(name):
@@ -91,6 +99,12 @@ class ContainerBind:
 
     def create_container(self, image, command=None, **kwargs):
         """Create and return running container instance."""
+        if net_name := settings.container.network:
+            if not self.get_network_by_attrs({"name": net_name}):
+                raise UserError(
+                    f"Network '{settings.container.network}' not found on container host."
+                )
+            kwargs["networks"] = {net_name: {"NetworkId": net_name}}
         kwargs = self._sanitize_create_args(kwargs)
         return self.client.containers.create(image, command, **kwargs)
 
@@ -106,6 +120,12 @@ class ContainerBind:
     def pull_image(self, name):
         """Pull an image into the container host."""
         return self.client.images.pull(name)
+
+    def get_network_by_attrs(self, attr_dict):
+        """Return the first matching network that matches all attr_dict keys and values."""
+        for network in self.networks:
+            if all(network.attrs.get(k) == v for k, v in attr_dict.items()):
+                return network
 
     @staticmethod
     def get_logs(container):
@@ -144,8 +164,10 @@ class PodmanBind(ContainerBind):
         self._ClientClass = PodmanClient
         if self.host == "localhost":
             self.uri = "unix:///run/user/1000/podman/podman.sock"
-        else:
+        elif kwargs.get("port") == SSH_PORT:
             self.uri = "http+ssh://{username}@{host}:{port}/run/podman/podman.sock".format(**kwargs)
+        else:
+            self.uri = "tcp://{host}:{port}".format(**kwargs)
 
     def _sanitize_create_args(self, kwargs):
         from podman.domain.containers_create import CreateMixin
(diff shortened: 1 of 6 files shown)


broker/binds/containers.py
```diff
@@ -107,6 +119,12 @@ def pull_image(self, name):
         """Pull an image into the container host."""
         return self.client.images.pull(name)
 
+    def get_network_by_attrs(self, attr_dict):
+        """Return the first matching network that matches all attr_dict keys and values."""
+        for network in self.networks:
+            if all(network.attrs.get(k) == v for k, v in attr_dict.items()):
+                return network
```
omkarkhatavkar, line 126: in our case I have disabled the ipv4 interfaces even from the network, if we go though this logic it will work gets the same network. But in case someone creates another network then it will have ipv4 and ipv6 interface as well. @JacobCallahan Do you think that it will be good to pass the network name/id from the broker_settings file ? it will also eliminate to have iteration as well.

omkarkhatavkar (approved):
ACK, Just One Question Pending

broker/binds/containers.py
JacobCallahan: great suggestion. i've now added a new `network` setting under the container config that can accept a specific network name.

JacobCallahan:
@omkarkhatavkar i've tested changing the network locally and things look good, but can you validate it works on your end? You can check the network a container is associated with by this attribute chain.

```python
container_host._cont_inst.attrs["NetworkSettings"]["Networks"]
```

broker_settings.yaml.example
```diff
@@ -35,6 +35,7 @@ Container:
             host_password: "<plain text password>"
             host_port: None
             runtime: docker
+            network: null
```
omkarkhatavkar, line 38: this needs to be updated in the CI.

broker_settings.yaml.example
```diff
@@ -35,6 +35,7 @@ Container:
             host_password: "<plain text password>"
             host_port: None
```
omkarkhatavkar, line 36: We might have to add this port in CI, it is empty. In this case, it uses the default 22, leading to SSH over TCP.

omkarkhatavkar: may be export in the pipeline

broker/binds/containers.py
```diff
@@ -91,6 +99,12 @@ def image_info(self, name):
 
     def create_container(self, image, command=None, **kwargs):
         """Create and return running container instance."""
+        if net_name := settings.container.network:
```
rplevka, line 102: why not making this setting a list and treat it that way? You would add support for multiple networks easily

tests/providers/test_container.py
```diff
@@ -76,6 +93,12 @@ def test_host_creation(container_stub):
     assert host.hostname == "f37d3058317f"
 
 
+def test_ipv6_host_creation(container_stub):
+    settings.container.network = "podman2"
```
omkarkhatavkar, line 97: this should be podman1, nevermind

omkarkhatavkar (approved):
nice work, Thanks for adding this @JacobCallahan

broker/binds/containers.py
JacobCallahan: Added in the latest update

broker/binds/containers.py
```diff
@@ -91,6 +99,13 @@ def image_info(self, name):
...
+            net_dict = {}
+            for name in net_name.split(","):
+                if not self.get_network_by_attrs({"name": name}):
+                    raise UserError(f"Network '{name}' not found on container host.")
+                net_dict[name] = {"NetworkId": name}
```
rplevka, line 107: is `NetworkId` really a reference to a network name?
from cli:
```
rplevka@rplevka-thinkpadt14gen4:~$ podman network list
NETWORK ID    NAME        DRIVER
2f259bab93aa  podman      bridge
adb2ef274993  podman1     bridge
```

rplevka: update: according the docs, the endpoint accepts ids OR names, so it is fine.

broker/binds/containers.py
```diff
@@ -144,8 +165,10 @@ def __init__(self, **kwargs):
         self._ClientClass = PodmanClient
         if self.host == "localhost":
             self.uri = "unix:///run/user/1000/podman/podman.sock"
-        else:
+        elif kwargs.get("port") == SSH_PORT:
```
rplevka, line 168: wouldn't it be better to check for existing schema in the `{host}` bit, and only apply this default in case it is not explicitly specified? That would also support e.g. ssh on a non-standard port

broker/broker.py
```diff
@@ -113,6 +113,8 @@ def __init__(self, **kwargs):
     def _act(self, provider, method, checkout=False):
         """Perform a general action against a provider's method."""
         logger.debug(f"Resolving action {method} on provider {provider}.")
+        if provider.__name__ in self._kwargs:
+            self._kwargs[provider.__name__] = self._kwargs.pop(provider.__name__)
```
ogajduse, line 117: Can you please clarify why this code is needed here? 
```python
[ins] In [30]: mydict = {'foo': 'bar'}

[ins] In [31]: mydict['foo'] = mydict.pop('foo')

[ins] In [32]: mydict
Out[32]: {'foo': 'bar'}
```
This seems like an unnecessary step. Isn't it missing something?

Having a comment here would be nice.

broker/binds/containers.py
JacobCallahan: for now, this will be fine. if we run into issues with non-standard ssh ports in the future, then we can re-evaluate for sure.

broker/broker.py
JacobCallahan: great catch! this was originally implemented with a different argument `provider_instance` that I ultimately decided didn't fit into the previously established way of specifying an alternate instance. This section looked like this before an apparently too-late-in-the-day refactor!
```python
        if "provider_instance" in self._kwargs:
            self._kwargs[provider.__name__] = self._kwargs.pop("provider_instance")
```

PR #303, opened 2024-08-06
Fix Broker's handling of argument assignment

JacobCallahan:
Ian noticed an issue had when specifying argument values in broker's cli, specifically when using an `=` assignment instead of space-delimited. This is likely an issue with the way click bundles arguments in its ctx.args.


tpapaioa:
LGTM

PR #304, opened 2024-08-08
Quick fix for docker ssh connection

JacobCallahan:
This time correctly accounting for multiple connection options in the Docker bind, allowing ssh (which is what most people are currently using) to work.


JacobCallahan:
Quick merge since this was tested on a hot system.

PR #310, opened 2024-09-06
Fix incomplete file downloads in ssh2 sftp_read

tpapaioa:
This PR fixes a bug in the ssh2 / ssh2-python312 implementation of sftp_read, which was only saving the last chunk of file data locally, in cases where the sftp transfer involves multiple reads.


JacobCallahan (approved):
ACK, great catch!

PR #317, opened 2024-09-17
Introduction of broker config cli and much more!

JacobCallahan:
The initial change of this was to introduce the broker config cli. However, to do this, a number of supporting and related changes needed to be made.

The first of these is the new ConfigManager class that manages reading from and writing to the config file.
Since so much logic was implemented for this, it made more sense to roll in the import/init functionality, that was originally in the settings module, into the new class.

This also gave me the opportunity to separate the test config from the example config. This removal, and other changes in this were supported by ConfigManager's migration functionality.

To simplify the nested chunk notation, I moved provider instances out from a list to just being a nested dictionary. This actually simplified the instance logic quite a bit.

Related to config simlification was the separation of the host settings to a new ssh chunk of the config. This is because not all users use the host functionality Broker provides. A likely change with this release is going to be removing a default ssh backend from Broker's requirements.

While I was at it, I also switched the CLI over to rich-click as part of a larger effort to improve the CLI experience.

### DIFF
--- a/broker/commands.py
+++ b/broker/commands.py
@@ -4,17 +4,28 @@ from functools import wraps
 import signal
 import sys
 
-import click
 from logzero import logger
 from rich.console import Console
+from rich.syntax import Syntax
 from rich.table import Table
+import rich_click as click
 
 from broker import exceptions, helpers, settings
 from broker.broker import Broker
+from broker.config_manager import ConfigManager
 from broker.logger import LOG_LEVEL
 from broker.providers import PROVIDER_ACTIONS, PROVIDER_HELP, PROVIDERS
 
 signal.signal(signal.SIGINT, helpers.handle_keyboardinterrupt)
+CONSOLE = Console()  # rich console for pretty printing
+
+click.rich_click.SHOW_ARGUMENTS = True
+click.rich_click.COMMAND_GROUPS = {
+    "broker": [
+        {"name": "Core Actions", "commands": ["checkout", "checkin", "inventory"]},
+        {"name": "Extras", "commands": ["execute", "extend", "providers", "config"]},
+    ]
+}
 
 
 def loggedcli(group=None, *cli_args, **cli_kwargs):
@@ -47,7 +58,7 @@ def parse_labels(provider_labels):
     }
 
 
-class ExceptionHandler(click.Group):
+class ExceptionHandler(click.RichGroup):
     """Wraps click group to catch and handle raised exceptions."""
 
     def __call__(self, *args, **kwargs):
@@ -167,12 +178,8 @@ def populate_providers(click_group):
 def cli(version):
     """Command-line interface for interacting with providers."""
     if version:
-        from importlib.metadata import version
-
         from packaging.version import Version
         import requests
-
-        broker_version = version("broker")
 
         # Check against the latest version published to PyPi
         try:
@@ -181,19 +188,28 @@ def cli(version):
                     "version"
                 ]
             )
-            if latest_version > Version(broker_version):
+            if latest_version > Version(ConfigManager.version):
                 click.secho(
                     f"A newer version of broker is available: {latest_version}",
                     fg="yellow",
                 )
         except requests.exceptions.RequestException as err:
             logger.warning(f"Unable to check for latest version: {err}")
-        click.echo(f"Version: {broker_version}")
-        broker_directory = settings.BROKER_DIRECTORY.absolute()
-        click.echo(f"Broker Directory: {broker_directory}")
-        click.echo(f"Settings File: {settings.settings_path.absolute()}")
-        click.echo(f"Inventory File: {broker_directory}/inventory.yaml")
-        click.echo(f"Log File: {broker_directory}/logs/broker.log")
+
+        # Create a rich table
+        table = Table(title=f"Broker {ConfigManager.version}")
+
+        table.add_column("", justify="left", style="cyan", no_wrap=True)
+        table.add_column("Location", justify="left", style="magenta")
+
+        table.add_row("Broker Directory", str(settings.BROKER_DIRECTORY.absolute()))
+        table.add_row("Settings File", str(settings.settings_path.absolute()))
+        table.add_row("Inventory File", f"{settings.BROKER_DIRECTORY.absolute()}/inventory.yaml")
+        table.add_row("Log File", f"{settings.BROKER_DIRECTORY.absolute()}/logs/broker.log")
+
+        # Print the table
+        console = Console()
+        console.print(table)
 
 
 @loggedcli(context_settings={"allow_extra_args": True, "ignore_unknown_options": True})
@@ -288,7 +304,6 @@ def inventory(details, curated, sync, filter):
     inventory = helpers.load_inventory(filter=filter)
     helpers.emit({"inventory": inventory})
     if curated:
-        console = Console()
         table = Table(title="Host Inventory")
 
         table.add_column("Id", justify="left", style="cyan", no_wrap=True)
@@ -302,7 +317,7 @@ def inventory(details, curated, sync, filter):
                 str(host["id"]), host["host"], host["provider"], host["action"], host["os"]
             )
 
-        console.print(table)
+        CONSOLE.print(table)
         return
     for num, host in enumerate(inventory):
         if (display_name := host.get("hostname")) is None:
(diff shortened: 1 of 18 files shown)


broker/config_manager.py
```diff
@@ -0,0 +1,264 @@
...
+    def backup(self):
+        """Backup the current configuration file."""
+        logger.debug(
+            f"Backing up the configuration file to {self._settings_path.with_suffix('.bak')}"
+        )
+        self._settings_path.with_suffix(".bak").write_text(self._settings_path.read_text())
+
+    def restore(self):
+        """Restore the configuration file from a backup if it exists."""
+        logger.debug(
```
JacobCallahan, line 120: I'm debating whether this should be info level or not. There is a stronger case for this being info level than the message in backup.

broker/config_manager.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def get(self, chunk=None, curr_chunk=None):
+        """Get a chunk of Broker's config or the whole config."""
+        if not curr_chunk:
+            curr_chunk = self._cfg
+        if not chunk:
+            return curr_chunk
+        if C_SEP in chunk:
+            curr, chunk = chunk.split(C_SEP, 1)
+            # curr = int(curr) if curr.isdigit() else curr
```
JacobCallahan, line 147: This is a remnant from when I allowed traversing lists, like how instances are currently stored. Could keep it in for some future use or just get rid of these lines.

broker/config_manager.py
```diff
@@ -0,0 +1,264 @@
...
+            raw_data = self._import_config(GH_CFG, is_url=True)
+        if not raw_data:
+            raise exceptions.ConfigurationError(
+                f"Broker settings file not found at {self._settings_path.absolute()}."
+            )
+        chunk_data = self.get(chunk, yaml.safe_load(raw_data))
+        edited_chunk = self._interactive_edit(chunk_data)
+        self.update(chunk, edited_chunk)
+
+    def migrate(self):
```
JacobCallahan, line 215: i've considered whether we should allow people to force migrations for a target version.

broker/hosts.py
```diff
@@ -45,6 +46,11 @@ def __init__(self, **kwargs):
             ipv6 (bool): Whether or not to use IPv6. Defaults to False.
             ipv4_fallback (bool): Whether or not to fallback to IPv4 if IPv6 fails. Defaults to True.
         """
+        global SETTINGS_VALIDATED  # noqa: PLW0603
+        if not SETTINGS_VALIDATED:
+            logger.debug("Validating ssh settings")
+            settings.validators.validate(only="SSH")
+            SETTINGS_VALIDATED = True
```
JacobCallahan, line 53: delay validation until needed to avoid requiring the ssh config chunk.
this isn't currently important since every setting under ssh has a default value defined in its validator.

broker/hosts.py
```diff
@@ -110,6 +117,8 @@ def connect(
             ipv6 (bool): Whether or not to use IPv6. Defaults to False.
             ipv4_fallback (bool): Whether or not to fallback to IPv4 if IPv6 fails. Defaults to True.
         """
+        from broker.session import Session
```
JacobCallahan, line 120: delay import so we can potentially remove the requirement to have a ssh backend installed.

broker/settings.py
```diff
@@ -97,23 +37,36 @@ def init_settings_from_local_repo(settings_path, interactive=False):
...
+
+if cfg_manager._get_migrations() and not TEST_MODE:
+    if INTERACTIVE_MODE:
+        click.secho(
+            "Broker settings file has pending migrations.\n"
+            "Continuing without running the migrations may cause errors.",
+            fg="red",
+        )
+        if click.confirm("Would you like to run the migrations now?"):
+            cfg_manager.migrate()
```
JacobCallahan, line 54: right now this is a choice, but we could also just remove the choice and make it automatic.

JacobCallahan:
@tpapaioa @ogajduse I'd like to have this merged this week; any thoughts?

broker/config_manager.py
JacobCallahan: we can now force

ogajduse (commented):
This is a partial review, I went through a quarter of the patch and left some comments.

tpapaioa (approved):

broker/commands.py
```diff
@@ -167,33 +178,38 @@ def provider_cmd(ctx, *args, **kwargs):  # the actual subcommand
...
+        table.add_column("Location", justify="left", style="magenta")
+
+        table.add_row("Broker Directory", str(settings.BROKER_DIRECTORY.absolute()))
+        table.add_row("Settings File", str(settings.settings_path.absolute()))
+        table.add_row("Inventory File", f"{settings.BROKER_DIRECTORY.absolute()}/inventory.yaml")
+        table.add_row("Log File", f"{settings.BROKER_DIRECTORY.absolute()}/logs/broker.log")
+
+        # Print the table
+        console = Console()
+        console.print(table)
```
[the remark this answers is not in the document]
JacobCallahan: good catch!

broker/config_manager.py
```diff
@@ -0,0 +1,294 @@
...
+        if self._interactive_mode is None:
+            self._interactive_mode = False
+            # GitHub action context
+            if "GITHUB_WORKFLOW" not in os.environ:
+                # determine if we're being ran from a CLI
+                for frame in inspect.stack()[::-1]:
+                    if "/bin/broker" in frame.filename:
+                        self._interactive_mode = True
+                        break
+        return self._interactive_mode
```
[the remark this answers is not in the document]
JacobCallahan: `sys.flags.interactive` is more for when python is ran with `-i`, so I don't want to go with that. However, I never thought about using `sys.stdin.isatty()` and that's a great suggestion. So I'm going to replace the post-github check with that.

broker/config_manager.py
```diff
@@ -0,0 +1,294 @@
...
+
+    def _import_config(self, source, is_url=False):
+        """Initialize the broker settings file from a source."""
+        proceed = True
+        if self.interactive_mode:
+            try:
+                proceed = click.confirm(f"Get example file from {source}?")
+            except click.core.Abort:
+                # We're likely in a different non-interactive environment (container?)
+                self._interactive_mode = False
```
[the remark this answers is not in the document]
JacobCallahan: it will likely never get hit, but just in case....

broker/providers/ansible_tower.py
```diff
@@ -21,6 +22,8 @@
 from broker import helpers
 from broker.providers import Provider
 
+YAML().register_class(awxkit.utils.PseudoNamespace)
```
JacobCallahan, line 25: Actually, it's probably better to act on `helpers.yaml` here
```suggestion
helpers.yaml.register_class(awxkit.utils.PseudoNamespace)
```

broker/providers/ansible_tower.py
```diff
@@ -778,9 +781,9 @@ def release(self, name, broker_args=None):
...
-def awxkit_representer(dumper, data):
-    """In order to resolve awxkit objects, a custom representer is needed."""
-    return dumper.represent_dict(dict(data))
+# def awxkit_representer(dumper, data):
+#     """In order to resolve awxkit objects, a custom representer is needed."""
+#     return dumper.represent_dict(dict(data))
 
 
-yaml.add_representer(awxkit.utils.PseudoNamespace, awxkit_representer)
+# yaml.add_representer(awxkit.utils.PseudoNamespace, awxkit_representer)
```
ogajduse, line 789: Can this chunk be deleted? We can restore it anytime from a specific git ref.

broker/settings.py
```diff
@@ -6,88 +6,28 @@
...
-    # determine if we're being ran from a CLI
-    for frame in inspect.stack()[::-1]:
-        if "/bin/broker" in frame.filename:
-            INTERACTIVE_MODE = True
-            break
-
-
+INTERACTIVE_MODE = ConfigManager().interactive_mode
 BROKER_DIRECTORY = Path.home().joinpath(".broker")
+TEST_MODE = "pytest" in sys.modules
```
ogajduse, line 26: Did you test that `pytest` is not in `sys.modules` when broker is imported in robottelo or a similar project?

tests/test_config_manager.py
```diff
@@ -0,0 +1,62 @@
+"""Test file for broker.config_manager module."""
+from ruamel.yaml import YAML
+from broker.config_manager import ConfigManager, GH_CFG
+from broker.settings import settings_path
+
+
+yaml = YAML()
```
ogajduse, line 7: As you suggested in another module, this could be changed to `from broker.helpers import yaml`.

broker/config_manager.py
```diff
@@ -0,0 +1,288 @@
...
+    @property
+    def interactive_mode(self):
+        """Determine if Broker is running in interactive mode."""
+        if self._interactive_mode is None:
+            self._interactive_mode = False
+            # GitHub action context
+            if "GITHUB_WORKFLOW" not in os.environ:
+                # determine if we're being ran from a CLI
+                if sys.stdin.isatty():
+                    self._interactive_mode = True
```
ogajduse, line 67: Doesn't `sys.stdin.isatty()` return False if it is being run from GH Actions?
```suggestion
            # determine if we're being run from a CLI
            if sys.stdin.isatty():
                self._interactive_mode = True
```

broker/config_manager.py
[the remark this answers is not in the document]
ogajduse: I have created a new thread on the same topic. Let's discuss it there.

broker/config_manager.py
```diff
@@ -0,0 +1,288 @@
...
+        logger.info("Config migration complete.")
+
+    def validate(self, chunk, providers=None):
+        """Validate a top-level chunk of Broker's config."""
+        if chunk == "all":
+            all_settings = [prov for prov in providers if prov != "TestProvider"] + ["base", "ssh"]
+            for item in all_settings:
+                self.validate(item, providers)
+            return
+        chunk = chunk.split(".")[0] if "." in chunk else chunk
```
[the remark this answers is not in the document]
JacobCallahan: good catch!

broker/providers/ansible_tower.py
JacobCallahan: yep, definitely should have been

broker/settings.py
JacobCallahan: it almost certainly should trigger test mode, so we wouldn't automatically migrate while in a pytest session.

tests/test_config_manager.py
JacobCallahan: for this, we don't need to modify the shared yaml object, so using it directly here to load is fine.

broker/config_manager.py
JacobCallahan: unsure, but could test it by adding this to the workflow temporarily
```
python -c "import sys; print(f'{sys.stdin.isatty()=}')"
```

JacobCallahan: update: it is indeed false, based on that test.
```
sys.stdin.isatty()=False
```

JacobCallahan: and now completely removed that property in favor of a class attribute, thanks!

broker/settings.py
ogajduse: Let me rephrase, do we want to have `TEST_MODE` set to `True` when we run robottelo (Foreman) tests? If we don't, this logic needs to be changed.
`pytest --setup-plan tests/foreman/api/test_ping.py -s`
```python
[74] > /home/ogajduse/repos/broker/broker/settings.py(28)<module>()
-> TEST_MODE = "pytest" in sys.modules
   2 frames hidden (try 'help hidden_frames')
(Pdb++) l
 23  	from broker.exceptions import ConfigurationError
 24  	
 25  	INTERACTIVE_MODE = ConfigManager.interactive_mode
 26  	BROKER_DIRECTORY = Path.home().joinpath(".broker")
 27  	breakpoint()
 28  ->	TEST_MODE = "pytest" in sys.modules
 29  	
 30  	if TEST_MODE:  # when in test mode, don't use the real broker directory
 31  	    BROKER_DIRECTORY = Path("tests/data/")
 32  	elif "BROKER_DIRECTORY" in os.environ:
 33  	    envar_location = Path(os.environ["BROKER_DIRECTORY"])
(Pdb++) "pytest" in sys.modules
True
```

broker/session.py
```diff
@@ -19,7 +19,7 @@
 from broker.settings import settings
 
 SSH_BACKENDS = ("ssh2-python", "ssh2-python312", "ansible-pylibssh", "hussh")
-SSH_BACKEND = settings.SSH_BACKEND
+SSH_BACKEND = settings.BACKEND
```
tpapaioa, line 22:
```suggestion
SSH_BACKEND = settings.SSH.BACKEND
```

tpapaioa (changes requested):

broker/session.py
JacobCallahan: fixed

ogajduse (approved):

PR #319, opened 2024-09-18
Better handling for providers in concurrent actions

JacobCallahan:
--- First Commit ---
This change simplifies the way we add provider validators to dynaconf. Instead of attempting to add new validators when a class instance is created, we instead add it when the class itself is created. This means that it will happen once immediately when broker is loaded.

--- Second Commit ---
This change rewrites the way broker actions handle concurrency.
Previously, this was offloaded in a manner that was wasteful in the way
it handled pre-action setup.
Now, we don't go concurrent until we're ready to perform the action we
want.

### DIFF
--- a/broker/providers/__init__.py
+++ b/broker/providers/__init__.py
@@ -65,18 +65,22 @@ class ProviderMeta(ABCMeta):
             for attr, obj in attrs.items():
                 if attr == "provider_help":
                     # register the help options based on the function arguments
-                    for name, param in inspect.signature(obj).parameters.items():
-                        if name not in ("self", "kwargs"):
-                            # {name: (cls, is_flag)}
-                            PROVIDER_HELP[name] = (
+                    for _name, param in inspect.signature(obj).parameters.items():
+                        if _name not in ("self", "kwargs"):
+                            # {_name: (cls, is_flag)}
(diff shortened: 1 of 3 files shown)


broker/providers/__init__.py
```diff
@@ -65,18 +65,22 @@ def __new__(cls, name, bases, attrs):
...
-                            logger.debug(f"Registered help option {name} for provider {name}")
+                            logger.debug(f"Registered help option {_name} for provider {_name}")
                 elif hasattr(obj, "_as_action"):
                     for action in obj._as_action:
                         PROVIDER_ACTIONS[action] = (new_cls, attr)
                         logger.debug(f"Registered action {action} for provider {name}")
+            # register provider settings validators
+            new_validators = [v for v in attrs.get("_validators") if v not in settings.validators]
+            logger.debug(f"Adding new validators: {[v.names[0] for v in new_validators]}")
+            settings.validators.extend(new_validators)
```
JacobCallahan, line 83: We could theoretically simplify the loading logic further with this.
```suggestion
            if "_validators" in attrs:
                settings.validators.extend(attrs["_validators"])
```

ogajduse (approved):

tpapaioa (approved):

PR #320, opened 2024-09-20
Fix demux handler

tpapaioa:
Podman has changed the return value format for exec commands, when `demux=True`:

https://github.com/containers/podman-py/pull/410

This PR updates `from_duplexed_exec` to handle the new return format.

PRT results: /job/robottelo-pr-testing/8753

### DIFF
--- a/broker/helpers.py
+++ b/broker/helpers.py
@@ -21,7 +21,6 @@ from logzero import logger
 import yaml
 
 from broker import exceptions, logger as b_log, settings
-from broker.binds.containers import demux_output
 
 FilterTest = namedtuple("FilterTest", "haystack needle test")
 INVENTORY_LOCK = threading.Lock()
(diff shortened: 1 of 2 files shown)


JacobCallahan:
Since this is new to 5.2 (don't currently know exact version), can you update the pyproject.toml to require >=5.2?

tpapaioa:
> Since this is new to 5.2 (don't currently know exact version), can you update the pyproject.toml to require >=5.2?

Updated.

broker/session.py
```diff
@@ -79,10 +79,13 @@ def run(self, command, demux=True, **kwargs):
             # Containers don't handle pipes, redirects, etc well in a bare exec_run
             command = f"/bin/bash -c '{command}'"
         result = self._cont_inst._cont_inst.exec_run(command, **kwargs)
+        logger.info(f"tpapaioa ContainerSession result 1: {command=} {result=}")
```
JacobCallahan, line 82: Some small cleanup needed

JacobCallahan (changes requested):

broker/session.py
tpapaioa: Updated

JacobCallahan (approved):
ACK, nice!

PR #324, opened 2024-10-07
Rework inventory cli

JacobCallahan:
With the changes we've been making in 0.6 toward a better, more customizable and visually appealing, user experience. It's time that we revisited the inventory command.
The --curated view introduced in 0.5.x has now become the default inventory view. The original default view is now present in its new form and home with the --list flag.
The --details flag largely retains the functionality it did before. However, it involves much less processing than before.

One important piece added in this is the ability to customize the inventory fields on a per-user basis. The details on this customization are documented in detail in the supporting code.
It will also be important to document this in the wiki upon release.

Here's a screenshot of what the new changes look like (with the details view cut-off for security).
Screenshot from 2024-10-07 16-26-56

### DIFF
--- a/broker/commands.py
+++ b/broker/commands.py
@@ -17,7 +17,7 @@ from broker.logger import LOG_LEVEL
 from broker.providers import PROVIDER_ACTIONS, PROVIDER_HELP, PROVIDERS
 
 signal.signal(signal.SIGINT, helpers.handle_keyboardinterrupt)
-CONSOLE = Console()  # rich console for pretty printing
+CONSOLE = Console(no_color=settings.settings.less_colors)  # rich console for pretty printing
 
 click.rich_click.SHOW_ARGUMENTS = True
 click.rich_click.COMMAND_GROUPS = {
--- a/broker/commands.py
+++ b/broker/commands.py
@@ -285,49 +285,52 @@ def checkin(hosts, background, all_, sequential, filter):
 
 @loggedcli()
 @click.option("--details", is_flag=True, help="Display all host details")
-@click.option("--curated", is_flag=True, help="Display curated host details")
+@click.option("--list", "_list", is_flag=True, help="Display only hostnames and local ids")
 @click.option(
     "--sync",
     type=str,
     help="Class-style name of a supported broker provider. (AnsibleTower)",
 )
 @click.option("--filter", type=str, help="Display only what matches the specified filter")
-def inventory(details, curated, sync, filter):
-    """Get a list of all hosts you've checked out showing hostname and local id.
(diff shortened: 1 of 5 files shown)


broker/commands.py
```diff
@@ -285,49 +285,46 @@ def checkin(hosts, background, all_, sequential, filter):
...
-        # if we're filtering, then don't show an index.
-        # Otherwise, a user might perform an action on the incorrect (unfiltered) index.
-        index = f"{num}: " if filter is None else ""
-        if details:
-            logger.info(f"{index}{display_name}:\n{helpers.yaml_format(host)}")
-        else:
-            logger.info(f"{index}{display_name}")
+
+    inventory_fields = (
+        {"Host": "hostname | name | ip"} if _list else settings.settings.inventory_fields
```
JacobCallahan, line 313: The list fields are hard coded here, but we could also make them configurable..

broker/config_migrations/v0_6_0.py
```diff
@@ -60,6 +61,28 @@ def add_thread_limit(config_dict):
     return config_dict
 
 
+def add_inventory_fields(config_dict):
```
JacobCallahan, line 64: Need to add this to the call chain below.

broker/commands.py
```diff
@@ -285,49 +285,46 @@ def checkin(hosts, background, all_, sequential, filter):
...
+@click.option("--list", "_list", is_flag=True, help="Display only hostnames and local ids")
 @click.option(
     "--sync",
     type=str,
     help="Class-style name of a supported broker provider. (AnsibleTower)",
 )
 @click.option("--filter", type=str, help="Display only what matches the specified filter")
-def inventory(details, curated, sync, filter):
+def inventory(details, _list, sync, filter):
     """Get a list of all hosts you've checked out showing hostname and local id.
```
JacobCallahan, line 296: Update the description here

broker_settings.yaml.example
```diff
@@ -4,6 +4,13 @@ _version: 0.6.0
 logging:
     console_level: info
     file_level: debug
+# Customize the fields and values presented by `broker inventory`
+# Almost all field values should correspond to a field in your Broker inventory
+inventory_fields:
```
JacobCallahan, line 9: Should likely add a validator for this

JacobCallahan:
@ogajduse enjoy.... lol
The color in there is likely just from how my terminal is setup to handle bolded text.
Screenshot from 2024-10-08 09-33-49

tpapaioa (changes requested):

tpapaioa (approved):

PR #328, opened 2024-10-15
More fixes

JacobCallahan:
These include resolving a problem with interactive edits on mac not working. This was likely due to a quirk with the NamedTemporaryFile implementation. I insteda switched to manually creating a temporary file and that worked.

It also includes an early escape for when a user's inventory is empty.

### DIFF
--- a/broker/commands.py
+++ b/broker/commands.py
@@ -326,6 +326,9 @@ def inventory(details, _list, sync, filter):
         )
         for host in inventory
     ]
+    if not curated_host_info:
+        CONSOLE.print("No hosts found in inventory.")
+        return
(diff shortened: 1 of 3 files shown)


broker/commands.py
```diff
@@ -326,6 +326,9 @@ def inventory(details, _list, sync, filter):
         )
         for host in inventory
     ]
+    if not curated_host_info:
+        CONSOLE.print("No hosts found in inventory.")
+        return
```
pcreech, line 331: Confirmed that this resolves the inventory issues I saw.

```
$ broker inventory
No hosts found in inventory.
```

PR #329, opened 2024-10-15
Remove deprecated method for pulling AnsibleTower host facts

tpapaioa:
This PR:
- removes the old method of pulling host data from AAP, which is no longer used since https://github.com/SatelliteQE/broker/pull/261
- updates the test data for the mock workflow job, to reflect the newer method

### DIFF
--- a/broker/providers/ansible_tower.py
+++ b/broker/providers/ansible_tower.py
@@ -22,14 +22,14 @@ from broker import helpers
 from broker.providers import Provider
 
 
-def convert_psuedonamespaces(attr_dict):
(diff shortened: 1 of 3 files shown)


broker/providers/ansible_tower.py
```diff
@@ -22,14 +22,14 @@
 from broker.providers import Provider
 
 
-def convert_psuedonamespaces(attr_dict):
+def convert_pseudonamespaces(attr_dict):
```
JacobCallahan, line 25: how. dare. you.

PR #338, opened 2024-11-26
Fix for ruamel interactions with awxkit objects

JacobCallahan:
Cole encountered an issue where inventory syncs were failing because ruamel wasn't able to handle awxkit's Psuedonamespace objects. This simply adds a new representer for them, fixing the issue.


ColeHiggins2 (approved):
Tested locally and works great!

PR #341, opened 2025-01-13
Make AnsibleTower checkout/execute actions more resilient

JacobCallahan:
We've been seeing some issues with service interruptions in AAP under high load. While the jobs do complete successfully, awxkit bails when encountering the connection issue.
With this change, we simple enter a retry loop when monitoring job status.

### DIFF
--- a/broker/providers/ansible_tower.py
+++ b/broker/providers/ansible_tower.py
@@ -8,6 +8,7 @@ from urllib import parse as url_parser
 import click
 from dynaconf import Validator
 from logzero import logger
+from requests.exceptions import ConnectionError
 
 from broker import exceptions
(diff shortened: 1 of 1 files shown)


ogajduse (approved):

bherrin3:
1. I was running AT-01 at full/over-full chat and deployed another 200 or so VMs using this PR which put AT-01 on life-support.
2. I had to commute so I shutdown my machine with an active broker session
3. Made it to location and had to reconnect VPN
4. All instances were hung in AT-01 and I had to give AT-01 CPR by killing A LOT of jobs including these
5. Once the jobs were killed, my terminal noted and reported the job failures.

That seems pretty indicative that this works for intent.

bherrin3 (approved):

PR #348, opened 2025-01-27
pyproject.toml updates

### LINKED ISSUE #347: Remove the dependence on ssh2-python312
For the primary ssh-backend dependency before Broker 0.7, let's move back to ssh2-python.

Read through the parent issue #342 for more information.

JacobCallahan:
Now that ssh2-python has returned, let's switch the primary ssh-backend dependency to it. This primary dependency will be removed in Broker 0.7 along with the ssh2_python311 and ssh2_python312 optional dep definitions.

Also, with the merge of #346, we can fix how the license is attributed in the project metadata. This moves us in compliance with the newer PEP 639 standard.

Fixes #347

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,10 +9,10 @@ readme = "README.md"
 requires-python = ">=3.10"
 keywords = ["broker", "AnsibleTower", "docker", "podman", "beaker"]
 authors = [{ name = "Jacob J Callahan", email = "jacob.callahan05@gmail.com" }]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -22,14 +22,14 @@ classifiers = [
 dependencies = [
     "awxkit",
     "click",
-    "dynaconf<4.0.0",
+    "dynaconf>=3.1.6,<4.0.0",
```
JacobCallahan, line 25: this was needed because selective validation was added in 3.1.6.

PR #352, opened 2025-02-19
Update the way we're getting a container's hostname

JacobCallahan:
Previously, if a user set a hostname on a container, Broker wouldn't recognize that. This is because we just looked up the container's id. When a hostname isn't set at runtime, this is perfectly fine. However, when a user does set a hostname, this breaks what would be expected.

With this change, we have a more reliable way to solve both cases.

### DIFF
--- a/broker/providers/container.py
+++ b/broker/providers/container.py
@@ -22,7 +22,7 @@ def container_info(container_inst):
         "_broker_provider": "Container",
         "_broker_args": helpers.dict_from_paths(container_inst.attrs, attr_dict),
         "name": container_inst.name,
-        "hostname": container_inst.id[:12],
+        "hostname": container_inst.attrs["Config"].get("Hostname"),
         "image": container_inst.image.tags,
         "ports": container_inst.ports,
     }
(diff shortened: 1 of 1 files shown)


aadhikar:
LGTM!

JacobCallahan:
Tests currently failing due to: https://github.com/containers/podman-py/issues/514, which @jyejare has already resolved. Hopefully we will get a new build of podman-py soon. However, I won't block this PR until then.

Manual testing of the changes show it working both for @aadhikar and myself.

```
❯ broker checkout --container-host localhost/ubi9:latest --hostname second.test.hostname
[INFO 250219 09:52:39] Using provider Container to checkout
[INFO 250219 09:52:39] Host: second.test.hostname
```
and in the inventory
```
4:
  name: jake_812af936
  _broker_provider_instance: podman_ipv4
  type: host
  hostname: second.test.hostname
  _broker_provider: Container
  _broker_args:
    container_host: localhost/ubi9:latest
    hostname: second.test.hostname
  exposed_ports: {
```
and now on the container itself
```
[root@infra-podman-ipv4 ~]# podman exec -it jake_812af936 /bin/bash
[root@second ~]# hostname
second.test.hostname
```

jyejare:
@JacobCallahan @aadhikar The podman-py new version is released with fix. I have rerun all the checks for this PR.

jyejare (approved):

PR #365, opened 2025-04-14
Add help entry describing how to chose specific AnsibleTower instance

jpasqualetto:
With this change, `broker checkout --help `  will show that `--AnsibleTower <instance> ` is a valid option to be used.


JacobCallahan (changes requested):
Instead of adding it as an explicit option, add the help text to the docstring. This pattern applies to all providers Broker supports, so the help text should reflect that.
Something along the lines of:
```
You can also checkout against a non-default provider instance
e.g. broker checkout ... --AnsibleTower <instance name>
```

PR #366, opened 2025-04-17
Allow checkout errors to propagate correctly

JacobCallahan:
This change is primarily based around propagating errors in a more correct way. Previously, we stripped them out during host construction, which was masked due to the errors still being logged on init.

Now, we keep the errors how they should be and raise the latest. This also combines a the loggedcli with the ExceptionHandler class in the cli to reduce redundancy.

Finally, we're not going to re-log the exception message as the command exits.

### DIFF
--- a/broker/broker.py
+++ b/broker/broker.py
@@ -16,6 +16,7 @@ Note:
     This module (or parent directory) should be used as the main entry point for the Broker API.
 
 """
+
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from contextlib import contextmanager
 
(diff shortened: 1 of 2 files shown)


JacobCallahan:
What this looks like now.

```shell
~/Pr/broker                                                                                                                                            
7❯ broker checkout --workflow deploy-satellite --deploy_sat_version 7.0
[INFO 250417 16:06:02] Using provider AnsibleTower to checkout
[INFO 250417 16:06:04] No inventory specified, Ansible Tower will use a default.
[INFO 250417 16:06:06] Waiting for job:
...
[ERROR 250417 16:07:17] JobExecutionError: AnsibleTower encountered the following error: [
      {
        "job": "satlab-tower-deploy-version-choice",
        "task": "Take Deploy Workflow inputs and output correct set-stats for the remainder of workflow",
        "reason": "No template found for passed arguments"
      }
    ]
[ERROR 250417 16:07:17] Command failed due to: JobExecutionError

~/Pr/broker                                                                                                                                           
7❯ echo $?
7
```

PR #370, opened 2025-05-08
Changes needed for AAP 2.5 compatibility

JacobCallahan:
This change makes it possible for users to interact with AAP 2.5+ without having to set the awxkit environment variable themselves.

This path currently is a bit heavy-handed, but may be simplified in the future.

### DIFF
--- a/broker/providers/ansible_tower.py
+++ b/broker/providers/ansible_tower.py
@@ -695,7 +695,8 @@ class AnsibleTower(Provider):
         job_api_url = url_parser.urljoin(self.url, str(job.url))
         # Need to change the url subject for job templates. If this increases, then come up with a better solution
         job_ui_url = url_parser.urljoin(
-            self.url, f"/{AAP_URL_PREFIX}/jobs/{'playbook' if subject == 'job_template' else subject}/{job_number}/output"
+            self.url,
+            f"/{AAP_URL_PREFIX}/jobs/{'playbook' if subject == 'job_template' else subject}/{job_number}/output",
         )
         helpers.emit(api_url=job_api_url, ui_url=job_ui_url)
         logger.info(f"Waiting for job: \nAPI: {job_api_url}\nUI: {job_ui_url}")
--- a/broker/providers/ansible_tower.py
+++ b/broker/providers/ansible_tower.py
@@ -3,17 +3,21 @@
 from functools import cache, cached_property
 import inspect
 import json
+import os
+import sys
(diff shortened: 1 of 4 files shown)


broker/providers/ansible_tower.py
```diff
@@ -22,13 +25,65 @@
...
+
+    This function can be called multiple times safely.
+    Returns True if reconfiguration was done, False otherwise.
+    """
+    # Exit immediately if we've already configured
+    if getattr(detect_and_reconfigure_for_aap25, "_configured", False):
+        return False
+
+    # Check explicit setting configuration first
+    if not (need_reconfigure := settings.ANSIBLETOWER.get("AAP_VERSION") == "2.5"):
```
rdrazny, line 46:
```suggestion
    if not (need_reconfigure := str(settings.ANSIBLETOWER.get("AAP_VERSION")) == "2.5"):
```
I'm hitting an issue with type conversion. If I set `aap_version: "2.5"` in the config file, despite having the number quoted, it's still handled as a float, not as a string. This causes the comparions with "2.5" string to be false. Either explicitly converting the version from config to string, or having ` == 2.5` without quotes fixes the issue. 

Besides, if the setting is present as `aap_version: ` (no version string defined) as produced by current config migration, the conditional check assumes that the AAP is not 2.5, and the API check below is never performed, resulting in a failure. The version check should be more explicit (check for null, check for the Y version?).
Maybe dropping that `return False` would be better, to always run the autodetection (unless explicit "2.5", which goes for reconfig right away? Just an idea.

rdrazny:
Tested the new version. Things work correctly, with or without `aap_version` set. If I set `aap_version` to an incorrect version for a specific AAP instance, broker fails (as expected).
I have found one issue, though. Broker produces correct UI links both for 2.4 and 2.5 now, with one exception - job templates run on 2.5.
Job run UI link for AAP 2.4 produced by broker:  https://<aap_24_instance>/#/jobs/job_template/5819229/output
Job run UI link for AAP 2.5 produced by broker (this one does not work): https://<aap_25_instance>/execution/jobs/job_template/146214/output
Working job run UI link for AAP 2.5: https://<aap_25_instance>/execution/jobs/playbook/146214/output

AAP 2.5 uses `playbook` instead of `job_template` in links for job template runs.

JacobCallahan:
@rdrazny thanks, I've added that workaround with the latest update

tpapaioa (approved):

PR #374, opened 2025-05-22
Create temporary tokens when AnsibleTower users provide basic auth

JacobCallahan:
In AAP 2.5, token-based authentication results in significantly less errors. Because of this, we are going to attempt to create a temporary token, when a user only gives us their username and password.

### DIFF
--- a/broker/providers/ansible_tower.py
+++ b/broker/providers/ansible_tower.py
@@ -4,6 +4,8 @@ from functools import cache, cached_property
 import inspect
 import json
 import os
+import random
+import string
 import sys
 from urllib import parse as url_parser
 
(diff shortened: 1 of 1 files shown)


broker/providers/ansible_tower.py
```diff
@@ -161,6 +161,37 @@ def get_awxkit_and_uname(
...
+
+    temp_token_desc = None
+
+    # If no token was provided, try to create a temporary token
+    if not token and uname and pword:
+        helpers.emit(auth_type="password")
+        logger.warning(
+            "You should be using token-based authentication.\n"
+            "I will attempt to create and use a temporary token."
+        )
```
ogajduse, line 173: I would prefer broker to be impersonal. 
```suggestion
        logger.warning(
            "Token-based authentication should be used.\n"
            "An attempt will be made to create and use a temporary token."
        )
```

PR #379, opened 2025-06-24
feat: add OpenStack provider with comprehensive support

aadhikar:
- Add OpenStack provider with openstacksdk integration
- Support multiple auth methods: clouds.yaml, app credentials, username/password
- Implement smart resource resolution (name-to-UUID) for images/flavors/networks
- Add template-based and direct parameter configuration
- Include configurable timeouts and proper error handling
- Add comprehensive test suite and documentation
- Support all broker lifecycle operations (checkout, checkin, extend, inventory)


broker/providers/openstack.py
```diff
@@ -0,0 +1,519 @@
...
+        Validator("OPENSTACK.project_domain_name", default="Default"),
+        Validator("OPENSTACK.identity_api_version", default="3"),
+    ]
+    _checkout_options = [
+        click.option("--image", help="Image name or UUID for the VM"),
+        click.option("--flavor", help="Flavor for the VM"),
+        click.option("--network", help="Network UUID for the VM"),
+        click.option("--key-name", help="SSH key pair name"),
+        click.option("--template", help="Template name from configuration"),
+        click.option("--project", help="Project/tenant name or ID"),
```
JacobCallahan, line 40: these are all very generic names that could apply to multiple providers. how about pre-fixing them with `ostack` or another appropriate name aside from `os`, which could be misinterpreted to mean operating system.

broker/providers/openstack.py
```diff
@@ -0,0 +1,519 @@
...
+        )
+
+        # Ensure instance_id is set if it's available in broker_args
+        if broker_args and "instance_id" in broker_args and not hasattr(host_inst, "instance_id"):
+            host_inst.instance_id = broker_args["instance_id"]
+
+    def provider_help(self, **kwargs):
+        """Print useful information from the OpenStack provider."""
+        # This method should not return anything, just log information
+        logger.info("OpenStack provider configured")
```
JacobCallahan, line 164: I'd prefer this either be implemented, since it is pretty core functionality or raise a `NotImplementedError` from `broker.exceptions`

broker/providers/openstack.py
```diff
@@ -0,0 +1,519 @@
...
+            "key_name": "default_key_name",
+        }
+
+        for param, default_key in default_mapping.items():
+            if param not in params:
+                params[param] = getattr(settings.OPENSTACK, default_key, None)
+
+        return params
+
+    @Provider.register_action("template", "image")
```
JacobCallahan, line 231: i think `template` and `image` are too generic. these should probably be prefixed like the above suggestions, unless there is a better name to use.

broker/providers/openstack.py
```diff
@@ -0,0 +1,519 @@
...
+                    raise exceptions.ProviderError(
+                        "OpenStack", f"Server {server.id} failed to build"
+                    )
+
+                time.sleep(5)  # Poll every 5 seconds
+            else:
+                raise exceptions.ProviderError(
+                    "OpenStack",
+                    f"Server {server.id} did not become active within {timeout} seconds",
+                )
```
JacobCallahan, line 306: broker has a `simple_retry` function in the `helpers` module that you may want to use instead.

docs/providers/openstack.md
```diff
@@ -0,0 +1,437 @@
+# OpenStack Provider
```
JacobCallahan, line 1: We don't really do docs within a specific directory. Instead, this project uses GitHub's wiki functionality to handle more user-facing documentation. I like a lot of the content you have here, but let's move it to the wiki after this has been merged.

tests/providers/test_openstack.py
```diff
@@ -0,0 +1,421 @@
...
+        "get": lambda key, default=None: getattr(mock_openstack_settings, key, default),
+    })
+
+    # Mock the settings object to have OPENSTACK attribute
+    mock_settings = MockStub({
+        "OPENSTACK": mock_openstack_settings
+    })
+
+    # Patch the settings import in the openstack provider module
+    monkeypatch.setattr("broker.providers.openstack.settings", mock_settings)
```
JacobCallahan, line 173: try to avoid monkey patching this and instead use dependency injection to pass along the settings object. This will also become important when we allow users to use non-global configs.

JacobCallahan:
@aadhikar I've updated your PR to point to the new 0.7 branch. I'll be collecting a number of large features there before putting together the full 0.7 release. You may have to deal with some code standards changes and merge conflicts.

broker/providers/openstack.py
aadhikar: @JacobCallahan Added support for exceptions in `helpers.py` that should cause immediate failure rather than retries. This lets the OpenStack provider fail quickly when servers enter the ERROR state instead of retrying for the full timeout period.

broker/helpers.py
```diff
@@ -562,13 +562,20 @@ def translate_timeout(timeout):
...
+):
     """Re(Try) a function given its args and kwargs up until a max timeout."""
     cmd_args = cmd_args if cmd_args else []
     cmd_kwargs = cmd_kwargs if cmd_kwargs else {}
+    terminal_exceptions = terminal_exceptions or ()
+
     try:
         return cmd(*cmd_args, **cmd_kwargs)
     except Exception as err:
+        if isinstance(err, terminal_exceptions):
```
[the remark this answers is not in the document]
JacobCallahan: this is the way

tests/conftest.py
```diff
@@ -78,7 +90,7 @@ def ensure_test_server_image(container_client):
         client.pull_image(TEST_SERVER_IMAGE)
 
 
-@pytest.fixture(scope="session", autouse=True)
+@pytest.fixture(scope="session")  # Removed autouse=True
```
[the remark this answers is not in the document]
aadhikar: Removed `autouse=True` because OpenStack tests were failing due to unnecessary container setup trying to import podman/docker. Fixed by adding explicit `run_test_server` dependencies to SSH tests that need it. Good catch. This definitely could have caused more issues. @JacobCallahan, I don’t think any other tests require this change, but please double-check when you get a chance.

JacobCallahan (approved):
ACK. Great addition!

PR #386, opened 2025-07-25
Implement local settings isolation for Broker

JacobCallahan:
This change enables multiple Broker instances to operate with different configurations simultaneously, supporting use cases like testing multiple environments or isolation while maintaining backward compatibility with existing global settings usage.

Features:
- Add settings_obj parameter to core classes (Broker, Provider, Host, Session) enabling custom settings injection
- Implement create_settings() function for generating isolated settings instances with custom configurations
- Add entry-points based SSH backend loading system replacing hardcoded imports
- Support component-specific settings files via component_settings_path()
- Enable multi_manager() to accept custom settings for isolated multi-host operations

Refactoring:
- Replace direct settings module access with configurable _settings attribute across all providers
- Modernize SSH session creation using entry-points discovery instead of conditional imports
- Consolidate settings validation and migration logic into create_settings function
- Update Host.connect() to use make_session() factory pattern with settings injection
- Streamline provider initialization to handle both global and custom settings objects

Configuration:
- Add SSH session and interactive shell entry-points in pyproject.toml
- Remove TEST_MODE environment variable dependency from settings initialization
- Enhance settings validators with safer defaults for SSH configuration
- Support network configuration in Container provider via settings object

Tests:
- Update all test fixtures to use isolated broker_settings instead of global settings
- Add comprehensive test data directory management with broker_data_dir fixture
- Improve parameterized test handling with proper monkeypatch usage
- Enhance provider test stubs to accept and use custom settings objects


broker/session.py
```diff
@@ -19,53 +20,101 @@
...
-
-logger.debug(f"{SSH_BACKEND=}")
-
-
-SSH_NOT_SUPPORTED_MSG = (
-    f"SSH backend {SSH_BACKEND!r} not supported.\nSupported ssh backends:\n{SSH_BACKENDS}"
-)
 SSH_NOT_INSTALLED_MSG = (
-    f"{SSH_BACKEND} is not installed.\n"
+    "{backend} is not installed.\n"
```
JacobCallahan, line 24:
```suggestion
    f"{backend} is not installed.\n"
```

JacobCallahan, line 24:
```suggestion
    f"{backend} is not installed.\n"
```

broker/session.py
```diff
@@ -19,53 +20,101 @@
...
     "ssh actions will not work.\n"
-    f"To use ssh, run 'pip install broker[{SSH_BACKEND}]'."
+    "To use ssh, run 'pip install broker[{backend}]'."
```
JacobCallahan, line 26:
```suggestion
   f "To use ssh, run 'pip install broker[{backend}]'."
```

broker/settings.py
```diff
@@ -81,24 +63,89 @@
...
+
+    return new_settings
+
+
+# Create the global settings object
+settings = create_settings()
+
+
+# Add helper function to determine settings file for a specific component
+def component_settings_path(component_name):
```
JacobCallahan, line 142: Is this really needed?

pyproject.toml
```diff
@@ -62,6 +62,22 @@ broker = "broker.commands:cli"
...
+ssh2-python = "broker.binds.ssh2:Session"
+ssh2-python312 = "broker.binds.ssh2:Session"
+hussh = "broker.binds.hussh:Session"
+
+[project.entry-points."broker.ssh.interactive_shell"]
+ansible-pylibssh = "broker.binds.pylibssh:InteractiveShell"
+ssh2-python = "broker.binds.ssh2:InteractiveShell"
+ssh2-python312 = "broker.binds.ssh2:InteractiveShell"
+
+[tool.setuptools]
```
JacobCallahan, line 76: Still needed after move to uv build backend?

JacobCallahan:
Still more to do when it comes to refactoring tests, but the general idea is here.

cc @RonnyPfannschmidt

broker/binds/containers.py
```diff
@@ -57,12 +56,15 @@ class ContainerBind:
...
-    def __init__(self, host=None, username=None, password=None, port=22, timeout=None):
+    def __init__(
+        self, host=None, username=None, password=None, port=22, timeout=None, broker_settings=None
+    ):
         self.host = host
         self.username = username
         self.password = password
         self.port = port
         self.timeout = timeout
+        self._settings = broker_settings
```
RonnyPfannschmidt, line 67: why is None valid here - it should not have a default of None

broker/binds/foreman.py
```diff
@@ -16,12 +16,17 @@ class ForemanBind:
...
     }
 
-    def __init__(self, **kwargs):
-        self.foreman_username = kwargs.get("foreman_username", settings.foreman.foreman_username)
-        self.foreman_password = kwargs.get("foreman_password", settings.foreman.foreman_password)
-        self.url = kwargs.get("url", settings.foreman.foreman_url)
-        self.prefix = kwargs.get("prefix", settings.foreman.name_prefix)
-        self.verify = kwargs.get("verify", settings.foreman.verify)
+    def __init__(self, broker_settings=None, **kwargs):
+        self._settings = broker_settings or settings
```
RonnyPfannschmidt, line 20: the import for the default should be lazy if possible

broker/binds/foreman.py
```diff
@@ -16,12 +16,17 @@ class ForemanBind:
...
+        self.foreman_username = kwargs.get(
+            "foreman_username", self._settings.foreman.foreman_username
+        )
+        self.foreman_password = kwargs.get(
+            "foreman_password", self._settings.foreman.foreman_password
+        )
+        self.url = kwargs.get("url", self._settings.foreman.foreman_url)
```
RonnyPfannschmidt, line 27: the pattern of kwargs.get(name, self,_settings.myname.attribute seems recurring

id almost recomment a metaprogramming abstaction where you define the attribute name at class declaration and add a item to tel lthe base lookup and the search key

broker/providers/__init__.py
```diff
@@ -108,12 +108,22 @@ class Provider(metaclass=ProviderMeta):
...
+        if broker_settings:
+            self._settings = broker_settings
+            # Add provider validators to custom settings
+            if validators := getattr(self.__class__, "_validators", None):
+                self._settings.validators.extend(validators)
+            logger.debug(
+                f"{cls_name} provider using custom settings object: {broker_settings.to_dict()}"
+            )
+        else:
+            self._settings = settings.dynaconf_clone()
```
RonnyPfannschmidt, line 124: im seeing a number of settings copy/clone/retain patterns - im starting to thing it may be necessary to add a helper thats consistently used

broker/providers/ansible_tower.py
```diff
@@ -116,9 +117,10 @@ def convert_pseudonamespaces(attr_dict):
     return out_dict
 
 
-def resilient_job_wait(job, timeout=None):
+def resilient_job_wait(job, timeout=None, broker_settings=None):
     """Wait for a job to complete. Retry on errors."""
-    timeout = timeout or settings.ANSIBLETOWER.workflow_timeout
+    _settings = broker_settings or settings
```
RonnyPfannschmidt, line 122: my personal opinion is that the settings variable should disappear in favour fo a helper that gets passed the passed in settings or none and returns a copy/ref to the settings

broker/providers/container.py
```diff
@@ -215,7 +224,9 @@ def construct_host(self, provider_params, host_classes, **kwargs):
             raise Exception(f"Could not determine container hostname:\n{cont_attrs}")
         name = cont_attrs["name"]
         logger.debug(f"hostname: {hostname}, name: {name}, host type: host")
-        host_inst = host_classes["host"](**{**kwargs, "hostname": hostname, "name": name})
+        host_inst = host_classes["host"](
+            **{**kwargs, "hostname": hostname, "name": name, "broker_settings": self._settings}
```
RonnyPfannschmidt, line 228: why is this using multi level unpack, this can be jsut parameter unpacking, no need for the extra dict

broker/config_manager.py
```diff
@@ -47,6 +48,9 @@ def __init__(self, settings_path=None):
         if settings_path:
             if settings_path.exists():
                 self._cfg = yaml.load(self._settings_path)
+            elif os.environ.get("BROKER_NO_GLOBAL_CONFIG"):
```
RonnyPfannschmidt, line 51: this needs to be available as api as well

broker/settings.py
```diff
@@ -81,24 +63,76 @@
...
+            "No Broker settings file found. Using default settings.\n"
+            "Run 'broker config init' to create a settings file.",
+            fg="yellow",
+        )
+
+    return new_settings
+
+
+# Create the global settings object
+settings = create_settings()
```
RonnyPfannschmidt, line 138: for import speed and control i recommend making this one lazy

RonnyPfannschmidt (changes requested):
this looks like a nice starting point - id strongly recommend unifying the broker settings passover as well as the assign from config pattern

also consider passing sub-objects only the sub-config

broker/binds/foreman.py
JacobCallahan: definitely possible and made (almost) every instance of these lazy. however, this will require a re-think of how Provider-specific validators are handled. Right now they are added to the main settings object when a validator class is constructed. This may need to move to instantiation, but it's a bit late so I'll have to think it over later.

broker/binds/foreman.py
JacobCallahan: It's sometimes difficult to balance beauty vs readability or the balance between explicit and implicit behavior.

broker/config_manager.py
JacobCallahan: one of the troubles here that i'd need to solve to do this would be deferring the logger module's use of settings. can check that though

JacobCallahan:
significant number of updates in the latest push including the ability to do this.

Screenshot From 2025-08-14 17-05-42

PR #390, opened 2025-08-14
add BROKER_INTERACTIVE_MODE env var support

aadhikar:


broker/config_manager.py
```diff
@@ -40,6 +41,11 @@ class ConfigManager:
     """
 
     interactive_mode = sys.stdin.isatty()
+
+    if "BROKER_INTERACTIVE_MODE" in os.environ:
+        env_val = os.environ["BROKER_INTERACTIVE_MODE"].lower()
+        if env_val in ("false", "0", "no"):
+            interactive_mode = False
```
JacobCallahan, line 48: If we instead treat the environment variable like a flag, then we can greatly simplify the check.
```suggestion
interactive_mode = os.environ.get("BROKER_INTERACTIVE_MODE", sys.stdin.isatty())
```

JacobCallahan (changes requested):
I'm not sure how I feel about this proposal just yet. Also, this may not be needed in Broker 0.7

broker/config_manager.py
aadhikar: will make the change! BTW, can you backport this to older versions, so that I can use it?

aadhikar:
Made the fix, kinda reverted to the old way.

```In [1]: from iqe_broker.broker import ResourceManager, Provider
[I 250815 12:11:12 v0_6_12:20] Running config migrations for 0.6.12.
[D 250815 12:11:12 v0_6_12:10] Adding max_resilient_wait setting to AnsibleTower provider.
[D 250815 12:11:12 config_manager:114] Backing up the configuration file to /Users/aadhikar/Work/iqe_broker_env/iqe_broker_env/lib/python3.13/site-packages/iqe_broker/conf/broker_settings.bak
[I 250815 12:11:12 config_manager:238] Config migration complete.```

PR #412, opened 2025-10-21
Add Tox config, update CI, and refine interactive mode

JacobCallahan:
Configuration:
- Add a comprehensive `tox.toml` configuration file to standardize and manage various testing environments locally and in CI. This includes:
    - Environments for running unit tests across Python versions 3.10 through 3.14.
    - Dedicated environments for linting and code formatting using Ruff.
    - Specific environments for testing different SSH backends (hussh, paramiko, ssh2-python).
    - Functional test environments for containers, Beaker, and Satlab integrations.
    - Development, quick validation, and clean-up environments.
- Update the GitHub Actions workflow to align with Tox's Python version support, upgrading the test matrix to include Python 3.14 and removing Python 3.10.

Features:
- Introduce the `BROKER_NON_INTERACTIVE` environment variable, allowing explicit control to force the application into non-interactive mode. This is particularly useful for automated testing and continuous integration pipelines.

Fixes:
- Robustify the `ConfigManager`'s detection of interactive mode by:
    - Prioritizing the `BROKER_NON_INTERACTIVE` environment variable.
    - Ensuring non-interactive mode is correctly set when `sys.stdin.isatty()` is false or in environments where interactive input is not available.
- Correct an internal attribute reference from `_interactive_mode` to `self.interactive_mode` in `ConfigManager` to properly access the public attribute.
- Enhance the `ConfigManager.get` method's error handling to:
    - Catch `TypeError` in addition to `KeyError` when attempting to access configuration chunks, improving resilience.
    - Explicitly check for and raise an error if a configuration chunk's value is `None`, providing clearer diagnostics.

Tests:
- The new `tox.toml` establishes a consistent and comprehensive framework for running all project tests, encompassing unit tests, SSH backend-specific tests, and functional tests.
- Configure Tox environments to set `BROKER_NON_INTERACTIVE=1` by default, ensuring tests run reliably without requiring manual user input.

### DIFF
--- a/tox.toml
+++ b/tox.toml
@@ -1,0 +1,273 @@
+# Tox configuration for the broker project
+# https://tox.wiki/en/stable/config.html
+
+requires = ["tox>=4.20"]
+env_list = ["py310", "py311", "py312", "py313", "py314", "lint"]
+
+# Labels for grouping test environments
+[labels]
+test = ["py310", "py311", "py312", "py313", "py314"]
+ssh = ["ssh-hussh", "ssh-paramiko", "ssh2-python"]
+functional = ["func-containers", "func-beaker", "func-satlab"]
+all = [
+    "py310",
+    "py311",
+    "py312",
+    "py313",
+    "py314",
+    "lint",
+    "ssh-hussh",
+    "ssh-paramiko",
+    "ssh2-python",
+    "func-containers",
+    "func-beaker",
+    "func-satlab",
+]
+
+# Base configuration for all run environments
+[env_run_base]
+description = "Run unit tests with {base_python}"
+package = "wheel"
+deps = ["pytest>=8", "pytest-randomly", "docker", "pexpect"]
+extras = ["ansibletower", "podman", "openstack"]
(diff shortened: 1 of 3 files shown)


tox.toml
```diff
@@ -0,0 +1,317 @@
...
+        "{posargs}",
+    ],
+]
+
+# Linting environment
+[env.lint]
+description = "Run code quality checks with ruff"
+skip_install = false
+deps = ["ruff"]
+commands_pre = []
```
ogajduse: Could the empty `commands_pre` be removed from all the config sections?

tox.toml
```diff
@@ -0,0 +1,317 @@
...
+        "broker",
+        "config",
+        "init",
+        "--from",
+        "{tox_root}/tests/data/broker_settings.yaml",
+    ],
+]
+commands = [["pytest", "-v", "{tox_root}/tests/test_ssh.py", "{posargs}"]]
+
+[env.ssh-all]
```
ogajduse: Do we need this section? Can't we just run `tox run -m ssh` to test all SSH backends?

That would also remove the dependency on the platform-specific tool - bash.

PR #424, opened 2025-11-14
Enhance dangling host prompt with failure reason

JacobCallahan:
Features:
- Add an optional 'reason' parameter to `handle_dangling_hosts` to provide context for a host being considered dangling.
- Log the specific failure reason when the `dangling_behavior` is set to "prompt", providing users with more information about why a host needs attention.
- Pass the relevant failure message reason to `handle_dangling_hosts` when a job encounters a failure, ensuring the context is available.
- Improve the user experience by providing critical context during the interactive prompt for handling dangling hosts.

### DIFF
--- a/broker/providers/ansible_tower.py
+++ b/broker/providers/ansible_tower.py
@@ -558,7 +558,7 @@ class AnsibleTower(Provider):
                     hosts.append(vm_name)
         return list(set(hosts))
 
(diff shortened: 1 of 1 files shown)


ogajduse (approved):
LGTM code-wise. I'd like to see @bherrin3's feedback on this too,

PR #430, opened 2025-11-20
only convert paths to strings when calling hussh read/write

rmynar:
The source and destination paths have to be converted to strings as introduced in #367
Conversion needs to be done only when calling the scp/sftp_read/write hussh functions, not for _set_destination.

Converting paths to string in _set_destination caused malfunction when only source filename was given. It is expected that the destination file will be the same as source file, but 'None' is set instead.

### DIFF
--- a/broker/binds/hussh.py
+++ b/broker/binds/hussh.py
@@ -71,7 +71,7 @@ class Session:
 
     @staticmethod
     def _set_destination(source, destination):
-        dest = destination or source
+        dest = str(destination or source)
         if dest.endswith("/"):
             dest = dest + Path(source).name
         return dest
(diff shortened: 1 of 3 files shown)


lpramuk (changes requested):
Missing changes to ` _set_destination(source, destination)` 
Please add them

broker/binds/hussh.py
```diff
@@ -71,7 +71,7 @@ def __init__(self, **kwargs):
 
     @staticmethod
     def _set_destination(source, destination):
-        dest = destination or source
+        dest = str(destination or source)
         if dest.endswith("/"):
             dest = dest + Path(source).name
```
lpramuk: You still keep `source` unconverted to string. 
it can be required for `Path()` to convert it to string.

```suggestion
            dest = dest + Path(str(source)).name
```

lpramuk (changes requested):
`source` is kept unconverted. To maintain the same functionality we probably need to convert also source to string

lpramuk (approved):

PR #435, opened 2025-12-02
Improve broker_origin for containers

JacobCallahan:
This previously combined old and new logic to result not only in buggy behavior, but also a loss of valuable information. We'll simplify by just not modifying the origin information.


ColeHiggins2 (approved):
passing locally for me. Merge it!

PR #453, opened 2026-02-09
Prevent broker instances from reconfiguring logging

JacobCallahan:
Fixes:
- Resolves issues where additional broker instances, particularly in multi_manager contexts, would incorrectly reconfigure global logging settings.
- Ensures logging is configured once at application startup and remains consistent across all Broker instances, preventing configuration breakage.

Refactoring:
- Removes the conditional logging setup from the Broker constructor.
- Removes the unused import of `setup_logging` from `broker.logging`.

### DIFF
--- a/broker/broker.py
+++ b/broker/broker.py
@@ -23,7 +23,6 @@ import logging
 
 from broker import exceptions, helpers
 from broker.hosts import Host
-from broker.logging import setup_logging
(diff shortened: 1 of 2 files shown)


rmynar (approved):

JacobCallahan:
Added the test copilot asked for

PR #501, opened 2026-05-12
Enhance FileLock to reduce race conditions and add random jitter

JacobCallahan:
Features:
- Introduced a delay_start parameter that adds a random jitter (0.2s to 3.0s) before attempting to acquire a lock. This helps prevent high-concurrency collisions when multiple processes are triggered simultaneously.

Fixes:
- Improved lock acquisition atomicity by replacing the check-then-touch pattern with an exclusive file creation call (open with "x" mode). This eliminates a race condition where two processes could simultaneously verify the lock's absence and both proceed to create it.

Refactoring:
- Updated wait_file logic to use a more robust try-except block for handling FileExistsError.
- Modified the wait_file method signature to allow for optional timeout overrides.

Documentation:
- Updated the FileLock class docstring to provide usage examples for the new delay_start feature and clarify how the timeout interacts with the initial jitter.

### DIFF
--- a/broker/helpers/file_utils.py
+++ b/broker/helpers/file_utils.py
@@ -6,6 +6,7 @@ from io import BytesIO
 import json
 import logging
 from pathlib import Path
+import random
 import tarfile
 import time
 from uuid import uuid4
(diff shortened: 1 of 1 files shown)


tpapaioa (approved):
LGTM, and works in my testing of robottelo's `SharedResource`.

Issue #508, opened 2026-06-01
Race condition in config migration: concurrent processes corrupt broker_settings.yaml


ogajduse:
## Summary

When multiple processes (e.g. pytest-xdist workers) initialize broker simultaneously, they all detect pending config migrations and race to migrate the same `broker_settings.yaml` file. One process reads the file while another is mid-write, getting `None` from `yaml.load()`, which crashes `_get_migrations()`.

## Root Cause

`ConfigManager.__init__()` reads the config file without any file-level locking. When multiple processes start simultaneously:

1. All detect `_version: 0.7.0` and pending migration to `v0_8_3`
2. All run `migrate()` → `yaml.dump(working_config, self._settings_path)` concurrently (line 263)
3. Any process that calls `yaml.load()` on the file mid-write gets `None` (partial/empty file)
4. `_get_migrations()` crashes: `self._cfg.get("_version", ...)` on `None`

Additionally, `_get_migrations()` has no guard against `self._cfg` being `None`.

## Evidence from CI Logs

All 4 xdist workers start at the same second (`09:16:21`):

**gw1** (wins the race, migrates successfully):
```
09:16:21 - gw1 - broker.config_migrations.v0_8_3 - INFO - Running config migrations for 0.8.3.
09:16:21 - gw1 - broker.config_manager - DEBUG - Backing up the configuration file to broker_settings.bak
09:16:21 - gw1 - broker.config_manager - INFO - Config migration complete.
```

**gw2, gw3** also migrate (redundant but succeed):
```
09:16:21 - gw2 - broker.config_manager - INFO - Config migration complete.
09:16:21 - gw3 - broker.config_manager - INFO - Config migration complete.
```

**gw0** (reads mid-write, crashes):
```
broker/config_manager.py:112: in _get_migrations
    config_version = Version(self._cfg.get("_version", "0.0.0"))
E   AttributeError: 'NoneType' object has no attribute 'get'
```

This cascades — gw0 fails on **every subsequent test** since `_SettingsProxy._ensure_settings()` retries `create_settings(perform_migrations=True)` each time, but the file was already migrated to 0.8.3, so the migration code no longer triggers and `yaml.load()` succeeds from that point... except that `_ensure_settings` already set `self._settings = None` context and the error propagates.

## Suggested Fix

Two issues need fixing:

1. **Guard `_get_migrations()` against `None` config**:
```python
def _get_migrations(self, force_version=None):
    if not self._cfg:
        return []
[...]

Issue #507, opened 2026-06-01
Race condition: inventory file corruption with concurrent multi-process access (e.g. pytest-xdist)


ogajduse:
## Summary

`update_inventory()` uses `threading.Lock()` (`INVENTORY_LOCK`) which only protects threads **within the same process**. When multiple processes access the inventory file concurrently (e.g. pytest-xdist workers), the YAML file gets corrupted mid-read/write, causing `ruamel.yaml.parser.ParserError`.

`FileLock` exists in `broker/helpers/file_utils.py` and is designed for cross-process locking, but it is **not used for inventory operations**.

## Observed Failure

Multiple xdist workers (gw0, gw2) simultaneously reading/writing `inventory.yaml` produce:

```
ruamel.yaml.parser.ParserError: while parsing a block collection
  in "inventory.yaml", line 1, column 1
expected <block end>, but found '?'
  in "inventory.yaml", line 35, column 1
```

Two scenarios observed:
1. **Read-during-write**: `load_inventory()` reads the file while another process is in the middle of writing it via `update_inventory()` — YAML is incomplete/corrupt.
2. **Write-during-write**: Two processes both call `update_inventory()` simultaneously — `threading.Lock()` provides no cross-process protection.

## Call Paths Affected

- `Broker.from_inventory()` → `load_inventory()` → `load_file()` — **no lock at all**
- `Broker.checkin()` → `update_inventory()` — protected by `threading.Lock()` only (ineffective cross-process)

## Suggested Fix

Use `FileLock` (already exists in `broker/helpers/file_utils.py`) for both `load_inventory()` and `update_inventory()`:

```python
from broker.helpers.file_utils import FileLock

def load_inventory(filter=None):
    from broker.settings import inventory_path
    with FileLock(inventory_path):
        inv_data = load_file(inventory_path, warn=False)
    if inv_data and filter:
        inv_data = eval_filter(inv_data, filter)
    return inv_data or []

def update_inventory(add=None, remove=None):
    from broker.settings import inventory_path
    ...
    with FileLock(inventory_path):
        inv_data = load_inventory()
        # ... read-modify-write cycle fully under file lock
        yaml.dump(inv_data, inventory_path)
```

This ensures the entire read-modify-write cycle is atomic across processes.

## Environment

- Broker version: 0.8.4
- Python: 3.12
- pytest-xdist with multiple workers (gw0, gw2)
- Observed in CI: InteroperabilityTesting-rhel9-sat-6.18-server

[...]

Issue #395, opened 2025-09-18
convert timestamps to a human-readable format


JacobCallahan:
I'm not sure if this should happen when ingesting the data from AAP or when displaying the inventory entries.

```
10:
  remove_time: '432000'
  shutdown_time: '1765804090'
  name: bherring-AAP-MR1384
  type: host
  hostname:
  ip: ...
  tower_inventory: osp-rhos01-satellite-users
  _broker_provider: AnsibleTower
  _broker_provider_instance: production
  _broker_args: {}
```

PR #510, opened 2026-06-09
Fix file write race conditions and add humanized timestamp display

JacobCallahan:
Addresses concurrent file access issues that caused YAML corruption and crashes when multiple processes (pytest-xdist workers, concurrent CLI commands) accessed shared files simultaneously.

Race condition fixes (Issues #507, #508):

- Enhanced FileLock with dual-layer locking for both inter-process (file-based) and intra-process (threading) protection
- Fixed inventory.yaml race condition by replacing threading.Lock with FileLock in load_inventory() and update_inventory()
- Fixed config file race condition by adding FileLock to all config_manager operations (init, backup, restore, update, migrate)
- Added defensive None-guard in _get_migrations() to prevent crashes during concurrent migrations
- Fixed scenario inventory and imports manifest race conditions with FileLock protection
- Fixed emitter race condition by replacing threading.Lock with FileLock

The enhanced FileLock now maintains a class-level dict of threading locks per file path with full reentrancy support using acquisition counting. This ensures both same-process threads and different processes are properly synchronized, while allowing nested locks from the same thread (e.g., when migrate() calls backup()). Thread locks are acquired before file locks and released after for proper cleanup.

Humanized timestamp display (Issue #395):

- Added humanize_inventory_times() helper function to recursively format all *_time fields in inventory host dictionaries
- Updated inventory_fields_to_dict to auto-format *_time fields
- Modified 'broker inventory --details' to display human-readable timestamps instead of epoch values

### DIFF
--- a/broker/commands.py
+++ b/broker/commands.py
@@ -378,7 +378,8 @@ def inventory(details, _list, sync, filter):
     helpers.emit({"inventory": inventory})
     # details is handled differently than the normal and list views
     if details:
-        detailed = helpers.yaml_format(dict(enumerate(inventory)))
+        humanized = [helpers.humanize_inventory_times(host) for host in inventory]
+        detailed = helpers.yaml_format(dict(enumerate(humanized)))
         CONSOLE.print(Syntax(detailed, "yaml", background_color="default"))
         return
 
--- a/broker/config_manager.py
+++ b/broker/config_manager.py
@@ -47,12 +47,16 @@ class ConfigManager:
     no_global_config = False
 
     def __init__(self, settings_path=None):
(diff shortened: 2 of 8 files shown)


ogajduse:
The locking changes touch 6 files but the new tests only cover timestamp formatting. Could we add at least a couple of tests for the enhanced FileLock?

Suggestions:
1. **Reentrancy** — nested `with FileLock(same_file)` shouldn't deadlock
2. **Thread contention on inventory** — parallel `update_inventory` calls shouldn't lose entries

These two cover the core correctness claims of the PR (reentrancy + thread safety) and are cheap to write.

PR #529, opened 2026-07-06
Add version and snap to broker inventory

ColeHiggins2:
Okay here me out... I checkout A LOT of VMs lol, and I hate having to scroll through them all looking for my satellite version and snap number. I know we can individually add this into our broker settings but I would like to propose adding this into our defaults.

*This is a feature for satellite vms. Hosts currently show "unknown" in both the Version and Snap field.*

what do you think? @JacobCallahan

### DIFF
--- a/broker/config_migrations/v0_6_0.py
+++ b/broker/config_migrations/v0_6_0.py
@@ -86,6 +86,8 @@ def add_inventory_fields(config_dict):
         "Provider": "_broker_provider",
         "Action": "$action",
         "OS": "os_distribution os_distribution_version",
+        "Version": "_broker_args.deploy_sat_version",
+        "Snap": "_broker_args.deploy_snap_version",
     }
     config_dict["inventory_list_vars"] = "hostname | name"
     return config_dict
--- a/broker/config_migrations/v0_8_10.py
+++ b/broker/config_migrations/v0_8_10.py
@@ -1,0 +1,28 @@
+"""Config migrations for versions older than 0.8.10 to 0.8.10."""
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+TO_VERSION = "0.8.10"
+
+
+def add_version_snap_inventory_fields(config_dict):
+    """Add Version and Snap to inventory_fields if not already present."""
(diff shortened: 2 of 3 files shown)


JacobCallahan:
These are good for the specific use case you mentioned. However, Broker's default values for these fields shouldn't be specific to the AnsibleTower provider, and more specifically our implementation of that provider.

Definitely share this with people in our group, but I can't add it as default.

ogajduse:
Nice quality-of-life win - I get the pain of scrolling through hosts looking for version/snap.

One thing worth discussing before this lands as a default: Satellite Engineering is broker's biggest user, but not its only one. Folks running the `AnsibleTower` provider against their own AAP instance won't have these `_broker_args` keys, so shipping a migration that adds `deploy_sat_version`/`deploy_snap_version` into everyone's `inventory_fields` on upgrade bakes a Satellite-specific convention into a shared tool's defaults.

Alternative: keep broker's defaults generic, and have Satellite Engineering maintain this in one shared config that composes on top of broker's defaults. Dynaconf already supports this via `includes` (https://www.dynaconf.com/configuration/?h=includes#includes), and I confirmed it works with broker's current settings loading as-is, no code change needed:

```yaml
# ~/.broker/broker_settings.yaml
dynaconf_include:
  - /path/to/shared/satqe_broker_settings.yaml
```
```yaml
# satqe_broker_settings.yaml
dynaconf_merge: true
inventory_fields:
  Version: _broker_args.deploy_sat_version
  Snap: _broker_args.deploy_snap_version
```

Or skip per-user file edits entirely, since broker already sets `ENVVAR_PREFIX_FOR_DYNACONF="BROKER"`:

```bash
export BROKER_INCLUDES_FOR_DYNACONF=/path/to/shared/satqe_broker_settings.yaml
```

Both merge cleanly into `inventory_fields` alongside broker's generic defaults - tested locally. Same "Version"/"Snap" columns for the whole team, from one shared file, without changing what every other broker user gets out of the box.

ColeHiggins2:
@JacobCallahan So we should close this?

PR #537, opened 2026-07-20
Fix AttributeError in Host.to_dict() for directly-constructed hosts

lpramuk:
## Summary
- `Host.to_dict()` raises `AttributeError` when called on a host constructed directly via `Host(hostname=...)` rather than checked out through Broker, because `_prov_inst` is not set in that case
- Fixed by using `getattr(self, "_prov_inst", None)` with a `None` fallback, consistent with how `name` is already handled in the same dict literal

## Test plan
- [ ] Construct a `Host` directly and call `to_dict()` — should return `{"_broker_provider_instance": None, ...}` without raising
- [ ] Construct a `Host` via Broker checkout and call `to_dict()` — existing behaviour unchanged

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/broker/hosts.py
+++ b/broker/hosts.py
@@ -209,7 +209,7 @@ class Host:
         """
         ret_dict = {
             "name": getattr(self, "name", None),
-            "_broker_provider_instance": self._prov_inst.instance,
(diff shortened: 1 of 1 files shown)


JacobCallahan (approved):
Nice!

PR #569, opened 2026-08-24
Overwrite compute attributes

knoppi:
We recently started using the broker provider `foreman` for deploying both to vmware as well as proxmox. Compute attributes differ. This PR allows to adjust them but continues providing suitable values for provisioning to VMware.

For this specific case we needed to pass structured data:

```
compute_attributes:
  node_id: <our hypervisor>
  start_after_create: 1
```

The second commit allows to find a correct representation in the inventory.

### DIFF
--- a/broker/helpers/inventory.py
+++ b/broker/helpers/inventory.py
@@ -49,8 +49,16 @@ def update_inventory(add=None, remove=None):
 
     :return: no return value
     """
+    from dynaconf.utils.boxing import DynaBox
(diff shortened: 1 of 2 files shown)


JacobCallahan (approved):
ACK. Good luck with further integrations! I'd be interested in seeing what you all are doing on the Foreman side to use it as a provider!

knoppi:
Thanks!

